PR #104, opened 2014-11-19
GetAlloc() function, and MustGetAlloc(), MustSelect() helper functions

fln:
I would like to add a new function GetAlloc() to sqlx package.

This function is useful when either 0 or 1 rows should be returned. The idea is to pass a pointer to the destination and this function will set it to nil or the data from DB. For pointer to be modified in the function user should pass a double pointer.

Usage example:
```go
var person *Person

db.GetAlloc(&person, "SELECT ... LIMIT 1")

if person == nil {
    log.Print("Got nothing")
} else {
    log.Print("Got person: ", person)
}
```

Without GetAlloc() user must check for sql.ErrNoRows if using Get() or use slices and Select().

These patches also contain MustGetAlloc() and MustSelect() helper functions to panic on error and tests for GetAlloc().

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -297,6 +297,11 @@ func (db *DB) Select(dest interface{}, query string, args ...interface{}) error 
 	return Select(db, dest, query, args...)
 }
 
+// Select using this DB, and panic on error.
+func (db *DB) MustSelect(dest interface{}, query string, args ...interface{}) {
+	MustSelect(db, dest, query, args...)
+}
+
 // Get using this DB.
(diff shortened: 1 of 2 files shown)


jmoiron:
I think `**` is a bit ugly in Go, and these changes would add to much surface area to the interface for too little gain.  I don't like the idea of returning nil pointers as an error when there is already an error interface and a standard way of doing it, even if it's a bit verbose.  I think in model layers where this is common it's trivial to develop these things on top of sqlx (even though I acknowledge that sqlx does have a lot of little trivial helpers in it already).

PR #113, opened 2014-12-24
Correct method for getting last insert id

bgreenlee:


jmoiron:
I think someone got to me in IRC and I fixed this.  Thanks!

Issue #139, opened 2015-05-15
reflectx changes


jmoiron:
I wanted to open an issue to discuss some ongoing work by @pkieltyka on reflectx.

pkieltyka:
The goal is to make reflectx reveal a lot more useful information about a struct and its tags.

Currently in master, reflectx ignores struct tag options which are pretty common and handy. As well the `fieldMap` which is the type that stores the information about the parsed structs only maps the name to a traversal (which is the index to a specific field in a nested structure). 

My branch expands on the info returned about fields in a struct without any interface changes to sqlx, just a very minor change. 

I've changed `type fieldMap map[string][]int` in favor of `fieldInfo` that provides:
* Index - this is the same as the path traversal in fieldMap)
* Path - the resolvable named path of a nested field map
* Name - the name from the parsed struct tag, ie. `db:"col,opt"`, name="col"
* Field - the reflect.StructField for reference
* Zero - the zero value of the Field, useful for being able to easily get default values
* Options - any key or key/value options in a struct tag
* Embedded - a boolean to inform if a nested field is an embedded struct

All of this stuff mostly comes down to how `Path`'s are resolved and composed. I've included a lot of examples in the tests, but here's a few...:

## Inline structs

```go
type Foo struct {
  A int
}

type Bar struct {
  Foo // `db:""` is implied for an embedded struct
  B   int
  C   int `db:"-"`
}

type Baz struct {
  A   int
  Bar `db:"Bar"`
}
m := NewMapperFunc("db", func(s string) string { return s })
z := Baz{}
z.A = 1
z.B = 2
z.C = 4
z.Bar.Foo.A = 3
```

The parsed mapping looks like:
```
Index: [0]       Path: "A"          Name: "A"
Index: [1]       Path: "Bar"        Name: "Bar"   Embedded: true
Index: [1 0]     Path: "Bar.Foo"    Name: "Foo"   Embedded: true
Index: [1 1]     Path: "Bar.B"      Name: "B"
Index: [1 0 0]   Path: "Bar.A"      Name: "A"
```

notice the "Bar.A" path which is actually a part of the embedded Foo struct. This allows you to have flat column structures (ie. stored in a table), but allows you to split up a table into more runtime structs that can easily be composed to either some namespace, or just inline.

Inline structs are the default for anonymously embedded structs, unless `db:"something"` is provided at which point, it will be namespaced by "something".

## Compose paths (aka columns) with struct tags

```go
type Employee struct {
[...]

Issue #131, opened 2015-04-07
better tools for join queries


jmoiron:
Having thought over `sqlx.In` and the way it interacts with `sqlx.BindNamed` for a while, I've started to also explore other ways that sqlx can help people with their queries.  Besides `IN` queries, one of the sticking points in SQL has been JOIN queries, especially when you've already "modeled" your tables appropriately.

@andybons asked me in IRC today if there was any better way of dealing with joins, and I said no, because the info you get from `Columns()` doesn't include [fully qualified names](http://stackoverflow.com/questions/268296/have-mysql-select-statement-return-fully-qualified-column-names-like-table-field).  However, that's just for `SELECT *`;  obviously,   I think something like this should be possible:

```go
type Person struct {
    Name string
}

type Place struct {
    Name string
}

// ...
func etc(db *sqlx.DB) {
    var pp struct {
        Person `db:"person,prefix=person."`
        Place  `db:",prefix=place."` // uses default Mapper
    }
    
    db.Get(&pp, `SELECT 
            person.name "person.name", 
            place.name "place.name" 
        FROM person, place LIMIT 1;`)
}
```

You could use `_` or some such so you wouldn't have to quote the names in the aliases as well.  If the repetition is bothersome, you could always:

```go
func qualify(name string) string {
    return fmt.Sprintf(`%v "%v"`, name, name)
}
```

Or even a simple parser similar to how `sqlx.In` and `sqlx.BindNamed` work:

```
query := qualify("SELECT q:person.name, q:place.name FROM person, place;")
```

The actual missing piece is a way to tell sqlx that an embedded struct should provide a prefix for its members.  I have to think this out a bit more;  the straightforward uses of this seem pretty useful, but multiple levels of prefixed embeds sounds complex and confusing.

tonyhb:
That would be pretty amazing.  It would be really helpful when composing structs together.

Maybe we could use prefixes on only composed, anonymous structs like this? http://play.golang.org/p/r7--RLCoAc

That way prefix won't litter your standard struct definitions.

jmoiron:
That's what I hinted at above with my anonymous struct typed `pp` which embedded two other structs.  It is even possible to pretty succinctly define reflexive joins (joins of a table with itself), which would typically defeat embedding:

```go
type Employee struct {
    Name string
    Id   int
    // BossId is an id into the employee table
    BossId sql.NullInt64
}

func etc(db *sqlx.DB) {
    type Boss Employee
    var employees []struct{
[...]

PR #140, opened 2015-05-15
reflectx ptrs branch

jmoiron:
This is the PR for #139 based on the ptrs branch.

### DIFF
--- a/reflectx/reflect.go
+++ b/reflectx/reflect.go
@@ -1,4 +1,4 @@
-// Package reflect implements extensions to the standard reflect lib suitable
+// Package reflectx implements extensions to the standard reflect lib suitable
 // for implementing marshaling and unmarshaling packages.  The main Mapper type
 // allows for Go-compatible named atribute access, including accessing embedded
 // struct attributes and the ability to use  functions and struct tags to
@@ -121,7 +121,7 @@ func (m *Mapper) FieldsByName(v reflect.Value, names []string) []reflect.Value {
 	return vals
 }
 
-// Traversals by name returns a slice of int slices which represent the struct
+// TraversalsByName returns a slice of int slices which represent the struct
 // traversals for each mapped name.  Panics if t is not a struct or Indirectable
 // to a struct.  Returns empty int slice for each name not found.
 func (m *Mapper) TraversalsByName(t reflect.Type, names []string) [][]int {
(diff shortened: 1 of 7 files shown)


reflectx/reflect.go
```diff
@@ -6,20 +6,63 @@
...
+	Embedded bool
+}
+
+type fields struct {
+	Index []*field
+	Paths map[string]*field
+	Names map[string]*field
+}
+
+func (f fields) GetByPath(path string) *field {
```
jmoiron: In what context would you use these paths?  If you are doing a marshaller or an unmarshaller, you'll probably be descending the object tree as you go, so you won't need to traverse in this manner.

reflectx/reflect.go
```diff
@@ -6,20 +6,63 @@
...
+	if fi, ok := f.Paths[path]; ok {
+		return fi
+	} else {
+		return nil
+	}
+}
+
+func (f fields) GetByTraversal(index []int) *field {
```
jmoiron: Might be easier as `GetByTraversal(indexes ...int)`

Shouldn't this be more or less the same as [FieldByIndex(index []int) Value](http://golang.org/src/reflect/value.go?s=22469:22515#L726)?   The only reason that `FieldByIndexes` is different from `reflect.FieldByIndex` is that it is designed to not panic if it encounters nil pointers to embedded structs, as it's designed to be a bit more convenient for unmarshallers.  That should really be documented somewhere.

reflectx/reflect.go
pkieltyka: thats true, they're more useful to help to build the fields.Paths map, but it could be useful to return the Path while traversing the index (tree). Ie. technically, reflectx could parse a json struct tag tree and returning the full path of a field may be useful there.

reflectx/reflect.go
pkieltyka: agreed

jmoiron:
I've made a few changes to some of the implementation:

* GetByPath was functionally equivalent to a blind-read from the map
* I added a btree to FieldInfo and the "root" type to the Map so you can traverse it like you would with reflect (except a lot faster);  this simplifies the Traversal code since you can use the Traversal directly instead of comparing it against every leaf node.

I was wondering, with the tree, is the Index/Traversal actually necessary?

Very close to merging this, probably will do so at GopherCon provided I can get the use-cases I need for https://github.com/jmoiron/sqlx/issues/131 working in some tests.

pkieltyka:
Looking good. I just pulled this branch and all of upper.io/db v2 tests still pass, and so do the tests of my API server!

I was searching the code of upperio and my API server and I didn't use GetByTraversal anywhere, I wrote it for completeness from the previous version. I think we should keep it for now with your new btree version.

Let me know if you run into issues with your test cases for #131.

schickling:
Is there anything else needed to get this merged into master?

arekkas:
This doesn't work with nested slices though, does it?

PR #145, opened 2015-06-09
Update ptrs branch

pkieltyka:
I've finished off the reflectx ptrs branch. As far as I'm concerned, this branch is ready for use (we've already been using the earlier versions in production at Pressly).

This PR updates:
1. Renaming field to FieldInfo
2. Renaming fields to StructMap
3. Yes, both FieldInfo and StructMap are exported since we return them from other exported functions. As well, libraries that do struct tag gymnastics can also interact with FieldInfo then
4. Implemented FieldMap method for completeness and added test case
5. Ran benchmarks of master vs ptrs for sqlx and reflectx, and they are almost identical
6. Added a test case and example in sqlx for a join query related to issue #131

### DIFF
--- a/reflectx/reflect.go
+++ b/reflectx/reflect.go
@@ -14,7 +14,7 @@ import (
 	"sync"
 )
 
-type Field struct {
+type FieldInfo struct {
 	Index    []int
 	Path     string
 	Field    reflect.StructField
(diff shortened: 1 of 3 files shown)


eexit:
Hello, 

I know this is old but this PR has some breaking changes... I haven't identified what it is yet but I'm been trying trying to find the latest PR merge compatible with my project and this after this PR, my tests break.

Unless there is undocumented BC, would you mind to have a look and see what could break in there?

Thanks!

PR #156, opened 2015-08-04
Refactor and optimize 'In'. Add benchmark.

nuss-justin:
Optimize indexing performance and allocations of sqlx.In, by using strings.IndexByte for finding '?' characters and precomputing the capacity for the new args slice. Also call reflectx.Deref directly to void allocating an error when an argument is not a slice.

Add a benchmark for calling 'In'.

Improvements on current go stable (1.4.2, linux/amd64):

```
benchmark       old ns/op     new ns/op     delta
BenchmarkIn     5913          1785          -69.81%

benchmark       old allocs     new allocs     delta
BenchmarkIn     32             13             -59.38%

benchmark       old bytes     new bytes     delta
BenchmarkIn     941           584           -37.94%
```

Same benchmark on tip (devel +3ae1704, linux/amd64):

```
benchmark       old ns/op     new ns/op     delta
BenchmarkIn     9461          1269          -86.59%

benchmark       old allocs     new allocs     delta
BenchmarkIn     28             13             -53.57%

benchmark       old bytes     new bytes     delta
BenchmarkIn     976           624           -36.07%
```

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"reflect"
 	"strconv"
+	"strings"
 )
 
 // Bindvar types supported by Rebind, BindMap and BindStruct.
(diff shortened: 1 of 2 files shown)


nuss-justin:
Added 2 more optimizations and updated benchmarks.

jmoiron:
Thanks so much!  Great change :)

PR #157, opened 2015-08-12
Add support for named pointers to structs

Pitmairen:
This makes it possible to use named struct pointers like this:

```go

type User struct {
    Name string
}
type Document struct {
    Title  string
    Owner  *User `db:"owner"`
    Author *User `db:"author"`
}
```


gregf:
:+1: I could really use this right now.

jmoiron:
This was an oversight, great commit & tests.  Thanks so much!

PR #159, opened 2015-08-14
Fix NumField panic when handling embedded structs

cortex:
This fixes a panic when dealing with embedded structs.

### DIFF
--- a/reflectx/reflect.go
+++ b/reflectx/reflect.go
@@ -270,10 +270,14 @@ func getMapping(t reflect.Type, tagName string, mapFunc, tagMapFunc func(string)
 		// pop the first item off of the queue
 		tq := queue[0]
 		queue = queue[1:]
-		tq.fi.Children = make([]*FieldInfo, tq.t.NumField())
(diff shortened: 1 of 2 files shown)


jmoiron:
Can you describe the situation that caused this?  I'd like to add a regression test.

cortex:
Sure. We're using UUIDs for our PKs, using m4rw3r's UUID lib, which defines UUID as:


```go
type UUID [16]byte
```
We then rename this type using embedded structs, to preserve Un/Marshalling:

```go
import	"github.com/m4rw3r/uuid"
// OfferID is a UUID
type OfferID struct {
	uuid.UUID
}
```

This worked fine in earlier versions of sqlx.

cortex:
I added a regression test. If you have suggestions for naming or folding into existing tests I'd be happy to do so.

cortex:
Any update in merging this? We are using a private fork for the time being because of this issue.

jmoiron:
Thanks so much, sorry for the delay in merging.

PR #161, opened 2015-08-20
Improved message for field not found

caarlos0:
I think that the current message is rather hard to understand,
  especially if the field that was not found is named "name"... I tried
  to improve the messages a bit, not sure if it is good though.

### DIFF
--- a/named.go
+++ b/named.go
@@ -151,7 +151,7 @@ func bindArgs(names []string, arg interface{}, m *reflectx.Mapper) ([]interface{
 	fields := m.TraversalsByName(v.Type(), names)
 	for i, t := range fields {
 		if len(t) == 0 {
-			return arglist, fmt.Errorf("could not find the field named %q in \"%#v\"", names[i], arg)
+			return arglist, fmt.Errorf("could not find the field named %q in %#v", names[i], arg)
 		}
 		val := reflectx.FieldByIndexesReadOnly(v, t)
 		arglist = append(arglist, val.Interface())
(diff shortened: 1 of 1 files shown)


named.go
```diff
@@ -151,7 +151,7 @@ func bindArgs(names []string, arg interface{}, m *reflectx.Mapper) ([]interface{
 	fields := m.TraversalsByName(v.Type(), names)
 	for i, t := range fields {
 		if len(t) == 0 {
-			return arglist, fmt.Errorf("could not find name %s in %#v", names[i], arg)
+			return arglist, fmt.Errorf("could not find the field named \"%s\" in \"%#v\"", names[i], arg)
```
elithrar: You can use `%q` here instead of quoting the verbs yourself, and/or use raw strings (backticks) to avoid having to escape the quotes.

caarlos0: Good catch, done.

caarlos0: Don't know how to do it with `%#v` though

elithrar: I would just use single quotes there - e.g. '%#v' - as %q also uses single
quotes.

caarlos0:
```go
package main

import "fmt"

func main() {
	fmt.Printf("blah %q", "asdsads")
}
```

```
$ go run main.go
blah "asdsads"
```

I think it's double quotes...

elithrar: Sorry, you're right. You probably don't need to quote the args list as %#v
is sufficiently obvious (being a syntax representation).

caarlos0: right, thanks!

caarlos0:
@elithrar done

caarlos0:
@elithrar ping

caarlos0:
@elithrar ping

robert-zaremba:
This PR is pretty obvious. What is blocking from merging it?

caarlos0:
well, I give up.

elithrar:
Sorry: I never saw the original pings and don't have merge rights here (I was reviewing as a fellow user).

PR #170, opened 2015-09-17
Add NamedStmt.Selectx

snikch:
First up, thanks for the great lib. As a kiwi it's cool to see the tests full of All Blacks too.

We're using named statements and a custom mapper, and couldn't use the handy `Select` method as the custom mapper would not get passed through. This fixes that. Let me know if this isn't a good solution.

Rugby Tax

### DIFF
--- a/named.go
+++ b/named.go
@@ -99,6 +99,17 @@ func (n *NamedStmt) Select(dest interface{}, arg interface{}) error {
 	return scanAll(rows, dest, false)
 }
 
+// Selectx using this NamedStmt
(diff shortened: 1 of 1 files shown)


jmoiron:
When I saw this PR at first I wondered why Select would need an x version, as it's already a custom extension.

I think this issue was also causing issue #181 where NamedStmts did not obey the "unsafe" flag.  It should be fixed now, ironically basically by replacing `Select` with the same code as this pull request.

Thanks for your contribution :)  `Select` should work as designed.

PR #173, opened 2015-09-27
Adding a sibling method to SliceScan to also get the columns as well.

StabbyCutyou:
I found myself wanting to use the SliceScan method for it's simplicity in dealing with the case where you don't know the incoming query or it's columns, and are willing to deal with reflection on your own as a result. Not for code in a critical path.

I ran into a case where I needed the column names, in order, along with the data in the columns. So I added this to my local fork, and it worked as expected. I tried to keep the code DRY, which is why I refactored the calls a little bit. If you have a suggestion for how you'd like to see it implemented or a better name, then I'm happy to do that work as well, but this is solving my problem at the moment.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -541,6 +541,11 @@ func (r *Rows) SliceScan() ([]interface{}, error) {
 	return SliceScan(r)
 }
 
+// SliceScanWithColumns using this Rows.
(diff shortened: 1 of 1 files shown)


StabbyCutyou:
So, turns out I am super dumb.

I didn't even realize that I could have just called Columns() my self from the calling code using SliceScan, after the first call to SliceScan.

Will refactor and close this in shame

PR #174, opened 2015-09-30
Add support for NamedQueryRow

proteneer:
We have functions:

```QueryRowx(query string, args ...interface{}) *Row```, which takes in a normal list of sql args, and returns a pointer to a single sqlx.Row

```NamedQuery(query string, arg interface{}) (*Rows, error)```, which takes in a single struct, and returns a pointer to a list of sqlx.Rows

This PR adds:
```NamedQueryRow(query string, arg interface{}) *Row```, which takes in a single struct, and returns a pointer to a single sqlx.Row

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -287,6 +287,12 @@ func (db *DB) NamedQuery(query string, arg interface{}) (*Rows, error) {
 	return NamedQuery(db, query, arg)
 }
 
+// NamedQueryRow using this DB.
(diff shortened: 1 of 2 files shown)


ottob:
Thanks for doing this but `NamedQueryRow` will panic if `NamedQuery` returns an error, due to rows being nil.

Fixed version: https://github.com/mindoktor/sqlx/commit/ac79aba637e9f513622bed998de2d0657ba35302

PR #179, opened 2015-10-13
Pretty printing for types.JsonText

mattbarton:
Without this, JsonText types print out as a byte array, e.g.: [123 34 110 97 109 101 34 58 32 34 77 97 116 116 32...]

With this, they print out as a normal string, e.g.: {"name": "Matt Barton", "gender": "male", "picture": "", ...


jmoiron:
Thanks!

PR #202, opened 2016-01-30
reflectx: update struct field walking for Go 1.6 changes

ianlancetaylor:
It's now necessary to check Anonymous as well as PkgPath to skip an
unexported field.  See http://tip.golang.org/doc/go1.6#reflect.

Fixes TestBindStruct when building with Go 1.6.


jmoiron:
Thanks!

PR #210, opened 2016-03-03
fix inconsistency between index.md and index.html

morenoh149:


jmoiron:
Thanks again :)  That's what not having an editor will do for you!

PR #212, opened 2016-03-04
Allow multiple tag names

snikch:
**Note** This PR isn't actually ready for merge since it updates the main package to use my fork

Hey there,

One of the use cases that I have for field names is the idea of generally using the `json` tag, but needing to override it on occasion for data that should be retrieved from the db, but not marshalled to json. By adding multiple tag support to `reflectx` I can save myself managing two tags.

For example, most of our data is encrypted with a per row nonce. We don't want to expose that nonce, but also don't want to expose the nonce via json. By using `json` tags for most fields, but using the `db` tag when it's available, I can easily provide this functionality.

``` go
type MyStruct struct {
  Name string `json:"name"`
  Nonce string `json:"-" db:"nonce"`
}

// My mapper
myDBClient.Mapper = reflectx.Mapper("db", "json")
```

Does this seem like something you'd like to support? If so we can talk about the implementation - my current one doesn't support multiple tags when using a mapper func - but a setter could be used to adjust or append to the tag names. It just seemed easiest to add a variadic parameter to the `reflectx.Mapper` function so that the api was backwards compatible.

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -7,7 +7,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/jmoiron/sqlx/reflectx"
+	"github.com/snikch/sqlx/reflectx"
 )
 
 // Bindvar types supported by Rebind, BindMap and BindStruct.
--- a/named.go
+++ b/named.go
@@ -243,7 +243,7 @@ func compileNamedQuery(qs []byte, bindType int) (query string, names []string, e
 			inName = true
 			name = []byte{}
 			// if we're in a name, and this is an allowed character, continue
-		} else if inName && (unicode.IsOneOf(allowedBindRunes, rune(b)) || b == '_') && i != last {
+		} else if inName && (unicode.IsOneOf(allowedBindRunes, rune(b)) || b == '_' || b == '.') && i != last {
 			// append the byte to the name if we are in a name and not on the last byte
 			name = append(name, b)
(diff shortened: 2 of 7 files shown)


elithrar:
Note that if you checkout a new branch within your GOPATH, make your
changes, and then `go install ./...` you don't have to rewrite the imports.

snikch:
@elithrar You mean in the main `github.com/jmoiron/sqlx` package? If so, I totally get that, but in the meantime I'm using this code in production so need to be able to retrieve from source control at the correct locations.

jmoiron:
My first impressions:

* this is a somewhat niche use case
* it's adding complexity for convenience rather than making something new possible
* the resulting code is less clear than just using a single tag consistently, even if that's repetitive

While the `reflectx.Mapper("db", "json")` semantics don't seem too difficult (prefer db, fallback to json if available), I think it hurts readability.  For example:

```go
type s struct {
    Y int `db:"-"`
    X int `json:"-"`
}
```

So I'm not sure I'm 100% on board, but I'm willing to be persuaded since you've had more experience with this problem and this particular solution!

snikch:
I understand your concerns, and it is definitely fitting a specific implementation into the current mapper model. An alternative would be to make a `sqlx.Mapper` interface, and then consumers are able to provide their own code as required?

PR #214, opened 2016-03-15
call NewMapperTagFunc's tagFunc before parsing name/options

jlburkhead:
May be I don't understand how `NewMapperTagFunc` is supposed to work, but it seems like the `tagMapFunc` needs to be called before [these lines](https://github.com/jlburkhead/sqlx/blob/c45bd30dde7197190548576fe3c5005d010dc75a/reflectx/reflect.go#L301) so that name and options are set for the field. Even if my understanding of how the `tagMapFunc` works is off, the test needs to be fixed since any func (or `nil`) works as the third arg to `NewMapperTagFunc` in `TestTagNameMapping`.


robert-zaremba:
I think you are right. `tag` is used to make find a column name, and this should be done before the name is set.

PR #222, opened 2016-05-12
Enable nested structs to work inside NamedQuery, NamedExec, etc.

cksmith:
This extends the work done in #131 so that fields from nested structs can be referenced in named queries (e.g. :place.name).  The only problem was that "." wasn't considered part of the field name.


robert-zaremba:
What is stopping from merging this?

PR #228, opened 2016-06-04
Minor spelling corrections

flimzy:


jmoiron:
https://golang.org/pkg/encoding/json/#Marshaler

Seems they're both valid.

flimzy:
Interesting... I noticed it was spelled both ways in comments in your package, and consulted a dictionary. If I had seen that first, I would have changed the other half. :)

jmoiron:
Apparently it is a UK/US split, as usual, with the US preferring single-l versions.  Despite being american myself, I think "canceled" and "marshaled" look weird and only used them because of their presence in the Go stdlib.  Thanks for making it consistent :)

PR #232, opened 2016-06-15
apply proper go attribute access rules in reflectx

### LINKED ISSUE #230: Scanning from enum no longer works
I recently updated sqlx and my queries no longer work when scanning postgres enum types into my Go structs.  I get this error: 

**error:** sql: Scan error on column index 2: unsupported Scan, storing driver.Value type []uint8 into type *stacks.StatusType"

**query:** ```db.Get(s, "SELECT * FROM stacks WHERE name = $1", name)```

**table:**
```sql
do $$
begin
	-- creates an enum type for status
	if not exists (select 1 from pg_type where typname = 'stack_status_type') then
		create type stack_status_type as enum ('beta', 'deprecated', 'public');
	end if;
end$$;


-- this table contains the supported Hooklift stacks metadata.
create table if not exists stacks (
	-- stack ID
	id                      uuid not null,
	-- stack's name
	name                    citext unique not null,
	-- status can be one of the following: beta, deprecated, public
	status                  stack_status_type not null default 'beta',
	-- timestamp of when the stack was created
	created_at              timestamptz not null default current_timestamp,
	-- timestamp of when the stack was last updated
	updated_at              timestamptz not null,

	primary key(id)
);
```

What am I missing?

jmoiron:
This fixes #230

There is a subtle change in behaviour here which I want
some time to confirm with the author of some of this code,
but I'm fairly certain that the new behaviour is:

* the advertised behaviour
* the old behaviour

### DIFF
--- a/reflectx/reflect.go
+++ b/reflectx/reflect.go
@@ -241,12 +241,6 @@ func methodName() string {
 	return f.Name()
 }
 
-type typeQueue struct {
-	t  reflect.Type
-	fi *FieldInfo
-	pp string // Parent path
-}
-
 // A copying append that creates a new slice each time.
 func apnd(is []int, i int) []int {
 	x := make([]int, len(is)+1)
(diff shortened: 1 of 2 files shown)


c4milo:
@jmoiron, any updates?

c4milo:
I just tested this patch and it doesn't seem to fix the problem.

PR #246, opened 2016-08-22
Make sqlx.scanAll() publicly available as sqlx.ScanAll()

chr4:
This is useful in some cases, e.g. when parsing multiple rows into an array of structs using
sqlx.NamedQuery().

```go
func GetWithNamedQuery(db sqlx.DB, target interface{}, query string, arg interface{}) (err error) {
    rows, err = db.NamedQuery(query, arg)
    if err != nil {
        return
    }

    // If something happens here, we want to make sure the rows are closed
    defer rows.Close()

    // This depends on sqlx.ScanAll() being an exported function.
    err = sqlx.ScanAll(rows, obj, false)
    return
}
```

The caller could then build (SQL injection secure) queries using something like this:

```go
// The id is mandatory
query := []string{`SELECT * FROM table WHERE id=:id`}

// Only filter for type if type was given
if where.Type != "" {
    query = append(query, `AND type=:type`)
}

// Only filter for state if state was given
if where.State != "" {
    query = append(query, `AND state=:state`)
}

// Order results
query = append(query, `ORDER BY created_at ASC`)
err = sql.NamedQuery(db, &contributions, strings.Join(query, " "), where)
```

I'm not sure whether this is the best option to build more dynamic queries, but I couldn't find a better option yet. Hints apprechiated.

### DIFF
--- a/named.go
+++ b/named.go
@@ -96,7 +96,7 @@ func (n *NamedStmt) Select(dest interface{}, arg interface{}) error {
 	}
 	// if something happens here, we want to make sure the rows are Closed
 	defer rows.Close()
-	return scanAll(rows, dest, false)
+	return ScanAll(rows, dest, false)
 }
 
 // Get using this NamedStmt
--- a/sqlx.go
+++ b/sqlx.go
@@ -633,7 +633,7 @@ func Select(q Queryer, dest interface{}, query string, args ...interface{}) erro
 	}
 	// if something happens here, we want to make sure the rows are Closed
 	defer rows.Close()
-	return scanAll(rows, dest, false)
+	return ScanAll(rows, dest, false)
 }
 
 // Get does a QueryRow using the provided Queryer, and scans the resulting row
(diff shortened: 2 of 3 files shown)


jmoiron:
I've felt that the solution to exposing more of this type of stuff or adding more `Named*` things to the DB object was to encourage use of `PrepareNamed`, which has `Select` et al.  I think that's sufficient for all use cases where you might want `ScanAll`.

chr4:
Thanks for your response!
I tried using `PrepareNamed()` in the example above, but I couldn't find a clean solution when trying to build more dynamic queries (like in the example above). Could you give me a hint/ an example how the problem could be solved using `PrepareNamed()` only?

The only/ cleanest thing I came up with is using `ScanAll()`, but I might have overlooked something.

jmoiron:
If you use `PrepareNamed`, then you can use `Select` to get the equivalent of `ScanAll`:

```go
stmt, err := db.PrepareNamed(`...`)
// handle err
var objs []MyObj
stmt.Select(&objs, where)
```

Please re-open if this isn't what you wanted!

chr4:
Thanks! This works great.

PR #247, opened 2016-08-27
JSONText & GzippedText are safe in 1.7

v4n:
1.7 contains this commit https://github.com/golang/go/commit/4e0cd1eeef419b221fda3dd3966be71095f0b4ce which resolves https://github.com/golang/go/issues/13905.


v4n:
@jmoiron let me know if you prefer any other kind of wording!

PR #251, opened 2016-09-14
better 'missing destination name' error message

mlitvin:
in now include not the type of the destination structure (as well as the name of the missing column)


jmoiron:
Thanks!

PR #254, opened 2016-10-13
Add type BitBool Scanner/Valuer for MySQL type BIT

RayfenWindspear:
Allows Go to read and write MySQL BIT(1) type as a bool. This allows you to use 1 bit for a MySQL boolean instead of 1 byte for the standard TINYINT. BitBool type itself is simply an alias for bool, so it can be used easily in Go code.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -107,3 +107,28 @@ func (j *JSONText) Unmarshal(v interface{}) error {
 func (j JSONText) String() string {
 	return string(j)
 }
+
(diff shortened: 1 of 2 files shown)


RayfenWindspear:
I've only tested this with MySQL, but it should work with any database types that get pulled in as a `[]byte`. Pretty sure it would at least also work with PostgreSQL.

jmoiron:
Thanks!

PR #119, opened 2015-01-11
handle nil values in sqlx/types.JsonText

pakohan:

PR #259, opened 2016-11-09
Add NullJSONText

Taik:
Forking off #119, this PR adds support for NullJSONText.

---
This change is [Reviewable](https://reviewable.io/reviews/jmoiron/sqlx/259)


kjk:
Would be great to have that. I just hit this problem where json type with null values in postgres db is not supported by sqls.

kjk:
In the mean-time: https://godoc.org/github.com/mc2soft/pq-types#JSONText.String supports null JSONText.

Issue #263, opened 2016-11-26
Regression bug? Infinite recursion for embedded structs


carbocation:
I think that the changes for #157 introduced a regression to a bug that was fixed for #60 . I'm seeing memory exhaustion when a struct contains another struct that embeds that struct type, which is seemingly caused by reflect descending into structs indefinitely since cc395bccac9fc05114b41b677fe5e663d8c52f6b (making 344b1e96d6e410a093b7bce40287731a49e253f6 the last good commit). 

Example structures:
```go
type Application struct {
    ID int64 `db:"id"`
    Entity *Entity `db:"entity_data"`
}

type Entity struct {
    Application

    Version string
}
```
Trying to structscan an Application no longer works after cc395bccac9fc05114b41b677fe5e663d8c52f6b .

I got here via bisecting:
```
git bisect start
# bad: [f71d5ee01769df469d115332106ef504ded071dc] Merge pull request #261 from gibson042/2016-11-placeholder-documentation
git bisect bad f71d5ee01769df469d115332106ef504ded071dc
# good: [debbfda33b4efc552406157658f49ab0c5ebeb31] Remove use of bytes.Count in sqlx.In
git bisect good debbfda33b4efc552406157658f49ab0c5ebeb31
# bad: [54aec3fd91a2b2129ffaca0d652b8a9223ee2d9e] Merge pull request #159 from cortex/master
git bisect bad 54aec3fd91a2b2129ffaca0d652b8a9223ee2d9e
# good: [cce5bd827a54704dd8da81bd16ddd20ce8316af4] add some documentation, simplify GetByPath a bit, add a comment about GetByTraversals implementation
git bisect good cce5bd827a54704dd8da81bd16ddd20ce8316af4
# bad: [3a92b1fd96032c4490fa618370fbf98ec69b680c] stop making a copy of the stmt for qstmt
git bisect bad 3a92b1fd96032c4490fa618370fbf98ec69b680c
# good: [344b1e96d6e410a093b7bce40287731a49e253f6] Merge pull request #140 from jmoiron/ptrs
git bisect good 344b1e96d6e410a093b7bce40287731a49e253f6
# bad: [ae682dc5c71d087ef97d506ec543320fdd7b0ff1] Merge pull request #157 from Pitmairen/master
git bisect bad ae682dc5c71d087ef97d506ec543320fdd7b0ff1
# bad: [cc395bccac9fc05114b41b677fe5e663d8c52f6b] Add support for named pointers to structs
[...]

PR #264, opened 2016-11-26
reflectx: don't mapping recursive field

tnextday:
the recursive struct will cause  `getMapping` loop forever.

e.g: 
```
type Person struct {
	Parent *Person
}
```

### DIFF
--- a/reflectx/reflect.go
+++ b/reflectx/reflect.go
@@ -330,10 +330,19 @@ func getMapping(t reflect.Type, tagName string, mapFunc, tagMapFunc mapf) *Struc
 	queue := []typeQueue{}
 	queue = append(queue, typeQueue{Deref(t), root, ""})
 
+QueueLoop:
 	for len(queue) != 0 {
 		// pop the first item off of the queue
 		tq := queue[0]
 		queue = queue[1:]
+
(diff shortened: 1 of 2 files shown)


tnextday:
https://github.com/jmoiron/sqlx/issues/263

jmoiron:
A more general solution would just skip any type already seen.  This is still succeptible to embed loops.  I'm not sure how the tests are passing because there should be a test for this already there.

tnextday:
Skip any type already seen may lose some field.

e.g:
```
type People struct {
	Name string
}

type Family struct {
	Wife *People
	Husband *People
}
```
-----------

```
type A1 struct {
	Person Person
}
type Person struct {
	Parent *Person
	A1 *A1
}
```

In this case, it will run more loops in my code, because I do not know how to compare structure and pointer types.

```
if tq.fi.Field.Type == p.Field.Type {
	continue QueueLoop
}
```

Sorry for my poor english.

jmoiron:
Thanks!  Pointer and non-pointer types are not the same and I think I'm ok with this, the second version will never be selected anyway.

tnextday:
Thank you!

PR #265, opened 2016-11-29
Information about the Get() errors included in the documentation

boreq:
I noticed that this (maybe obvious) information is missing from the documentation. Personally I had to check if `Get()` returns an error in case there are no results in the result set to be 100% sure. I am not sure if it fits better in the godoc or on the website but I think that is worth adding somewhere.


jmoiron:
Thanks!  `Select`/`Get` are supposed to mimic `Query`/`QueryRow` but I can see this was under-documented previously.

PR #267, opened 2016-12-06
Restore JSONText MarshalJSON behaviour

### LINKED ISSUE #266: JSONText broken
Until recently, `JSONText` was acting like a string, and since 6708aed04f34b30c834ad320fd3c2b8c61dd4c76, we have a lot of issues with this type.
See the following code:

```go
package main

import (
        "encoding/json"
        "fmt"
        "github.com/jmoiron/sqlx/types"
)

type X struct {
        Z types.JSONText
}

func main() {
        x := X{
                Z: types.JSONText("[]"),
        }
        bytes, _ := json.Marshal(x)
        fmt.Printf("%s\n", string(bytes))
}
```

before the commit, it was returning `{"Z":[]}` as expected, and after `{"Z":"W10="}`

gravis:
closes #266

When using `JSONText` in a struct, the field is not a pointer, and therefore the func is never called, leading to unexpected results.
This change works with `*JSONText`, as well as `JSONText`.

Thanks.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -60,11 +60,11 @@ type JSONText json.RawMessage
 var emptyJSON = JSONText("{}")
 
 // MarshalJSON returns the *j as the JSON encoding of j.
-func (j *JSONText) MarshalJSON() ([]byte, error) {
-	if len(*j) == 0 {
-		*j = emptyJSON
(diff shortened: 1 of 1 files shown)


jmoiron:
Cleaning this up, it was fixed in an earlier patch.

PR #268, opened 2016-12-08
scanAny: add nil check

pschultz:
The package-level function Get calls the interface method QueryRowx
which is expected to return *Row. Implementations of that interface
in another package can't possibly set the unexported rows field, so
it must be checked for nil before accessing it.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -717,6 +717,10 @@ func (r *Row) scanAny(dest interface{}, structOnly bool) error {
 	if r.err != nil {
 		return r.err
 	}
+	if r.rows == nil {
+		r.err = sql.ErrNoRows
+		return r.err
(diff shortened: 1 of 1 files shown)


pschultz:
Specifically, this blew up in my face in a test where Queryer is a mock

    import "github.com/stretchr/testify/mock"
    import "github.com/jmoiron/sqlx"
    
    block := func(mock.Arguments) { select{} }
    
    db := &TestQuerier{}
    db.On("QueryRowx", mock.Anything, mock.Anything).Run(block).Return(&sqlx.Row{})

    var n int
    sqlx.Get(db, &n, "SELECT 1")

jmoiron:
Thanks @pschultz!

PR #284, opened 2017-02-24
Add precisation about prepared statements issue with idle connections on MySQL

gm42:
This PR tries to achieve two goals:
* do not give incomplete/incorrect information about DB pools in Go (one should refer to official upstream documentation for that)
* add a precisation about an issue that happens when you have a very high idle connection pool and you prepare statements at application startup


gm42:
@jmoiron what about this one? close or keep?

PR #288, opened 2017-02-27
implement Tx.GetContext and Tx.SelectContext

### LINKED ISSUE #286: Missing tx methods with context
@wyattjoh @jmoiron 

- QueryRowxContext and GetContext
- QueryxContext and SelectContext

husio:
Missing context dependent methods implementations added to Tx structure.

Fix #286 

I could not find any tests for the `Tx` structure. Does this mean that it's not worth writing any for the `XxxContext` methods as well, or they are waiting for someone to add them?

### DIFF
--- a/sqlx_context.go
+++ b/sqlx_context.go
@@ -243,6 +243,36 @@ func (tx *Tx) MustExecContext(ctx context.Context, query string, args ...interfa
 	return MustExecContext(ctx, tx, query, args...)
 }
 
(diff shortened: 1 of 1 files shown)


jmoiron:
`GetContext` and `SelectContext` are confusing because the verbs suddenly sound like they are about the context!  I was wondering why `sql.Tx` didn't already implement this..  took me a while to get what this was.

PR #296, opened 2017-03-08
Add QueryxContext, QueryRowxContext and GetContext to sqlx.Tx

luna-duclos:
This adds QueryxContext, QueryRowxContext and GetContext functions to sqlx.Tx, without those, it is not possible to use cancellation or context-based tracing and logging.


luna-duclos:
I realized there is already a PR up for this, see #288

PR #305, opened 2017-04-12
Add rows.Err() calls to code in the docs

hypirion:
Hi, thanks for a good library!

Although the sqlx docs does refer to good sql documentation, many people tend to forget or don't always realise that you have to call `rows.Err()` after iterating via `rows.Next()`. I think adding them into the code examples – even without documenting why – should make more people aware that you have to do some error checking after a looping over rows.

### DIFF
--- a/index.html
+++ b/index.html
@@ -188,6 +188,7 @@ for rows.Next() {
     var telcode int
     err = rows.Scan(&country, &city, &telcode)
 }
+err = rows.Err()
 </pre>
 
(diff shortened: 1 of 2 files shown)


Taik:
Hi @hyPiRion, this is just for my own sanity--but can you explain a bit as to why that's required? Are you saying that the connection won't be fully released if `rows.Err()` isn't called?

hyPiRion:
@Taik: Oh, no problem. Calling `rows.Err()` doesn't have any impact on connections, so don't worry about leakage. When I wrote "have to call `rows.Err()`", I meant that for correctness' sake: You don't know whether or not your received all the rows, so you may return the wrong result.

PR #306, opened 2017-04-22
Return string type for JSONText.Value()

Taik:
In order to support `binary_parameters` flag in lib/pq, we need to explicitly return string type to signal lib/pq to send the field via text format (pointed out by cbandy at https://github.com/lib/pq/issues/528).

We've been testing this fork in production for a few days with pgbouncer transaction pooling mode, and so far things are working as expected.


dlsniper:

PR #327, opened 2017-06-16
Tests With docker-compose

RichardKnop:
I have added a platform agnostic way to run tests with docker-compose:

```sh
(docker-compose -f docker-compose.test.yml -p sqlx_ci up --build -d) && (docker logs -f sqlx_sut &) && (docker wait sqlx_sut)
```

This should make development easier as you can run tests from any machine with docker engine, you don't have to have a proper development setup locally.

Since the command is long, I have added a Makefile with alias to run tests:

```
make ci
```

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,0 +1,18 @@
+# Start from a Debian image with the latest version of Go installed
+# and a workspace (GOPATH) configured at /go.
+FROM golang
+
+# Set environment variables
+ENV PATH /go/bin:$PATH
+
+# Cd into the library code directory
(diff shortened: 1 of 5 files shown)


RichardKnop:
@jmoiron what do you think?

Also, the build badge/label on home page is broken right now, leading to 404 page at drone.io. Should we fix it?

jmoiron:
Hi Richard.  After a while away, I've added travis support for CI;  I honestly don't know enough about the docker ecosystem to comment on the rest.

tonyabracadabra (approved):

Issue #331, opened 2017-06-23
Proposal: MustScan


jakudlaty:
Hi
I think it will be nice to have Must* methods for querying data. Code could be more readable when used in some applications with panic recovery feature.

Instead of:
```
err := DB.Get(&date, "SELECT NOW();")
//error handling here
```
i could use
```
DB.MustGet(&date, "SELECT NOW();")
```

I can create a pull request for that.

jakudlaty:
I have created a simple pull request for that. #332

jmoiron:
Hi @jakudlaty!

I've shied away from adding these for a few reasons:

* adding MustGet without MustSelect feels like it's imballanced
* adding Must* to the whole API feels like too much
* the Must* methods in general are easy to write (see the bottom)
* the Must* methods in general should be avoided whenever possible

I feel a little guilty about some of the `Must` methods;  they enable things that are probably a bad idea.  As an example, I felt like `MustConnect` was somewhat analagous to `MustCompile`, but actually `MustConnect` is generally an _environment_ failure whereas `MustCompile` is likely to be a programmer error.  I also thought that `MustExec` would cut down on simple schema loading boiler plate, but again I think it just reinforces some poor error handling habits.  You're always better off initializing a connection where you can properly handle errors.

So I _kinda_ feel like the Must* methods were a mistake, and I don't want to add more to the interface.  The ones that are already there make sense in some situations, but not all.  In general, I suspect that such functions do have valid uses, but they are most likely to be used in places where not using them would be superior. 

Thankfully, you don't have to agree with me, and they aren't too hard to write in function form:

```go

func MustGet(q sqlx.Queryer, dest interface{}, query string, args ...interface{}) {
    if err := sqlx.Get(q, dest, query, args...); err != nil {
        panic(err)
    }
}
```

Thanks a lot for your contribution and I'm sorry I didn't reply sooner, but in the long run I don't think more `Must*` methods improve the already bloated API.

jakudlaty:
Thanks for your answer. I understand Your point of view, and now i know what was the reason for adding MustConnect and MustExec ;)

PR #332, opened 2017-06-26
Must* version of query fonctions

jakudlaty:


jmoiron:
I replied to this with reasons why I don't _really_ want to add this in [the proposal](https://github.com/jmoiron/sqlx/issues/331).

PR #337, opened 2017-07-12
Fix small typo

sevein:
I used redtape 1.2 to generate the HTML file.


jmoiron:
I think this makes you the second person to ever run redtape :laughing:   Thank you for the fix.

PR #340, opened 2017-07-18
[]byte should not be expanded by In

npiganeau:
[]byte is one of the driver.Value types, and should therefore not be expanded by `In`.

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -113,7 +113,7 @@ func In(query string, args ...interface{}) (string, []interface{}, error) {
 		v := reflect.ValueOf(arg)
 		t := reflectx.Deref(v.Type())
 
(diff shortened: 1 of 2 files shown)


bind.go
```diff
@@ -113,7 +113,7 @@ func In(query string, args ...interface{}) (string, []interface{}, error) {
 		v := reflect.ValueOf(arg)
 		t := reflectx.Deref(v.Type())
 
-		if t.Kind() == reflect.Slice {
+		if t.Kind() == reflect.Slice && t != reflect.TypeOf([]byte{}) {
```
nussjustin: Please add a comment explaining this exception and add a test case to TestIn in sqlx_test.go.

PR #341, opened 2017-07-21
compare NameMapper and origMapper properly

paulzhol:
https://golang.org/pkg/reflect/#Value:
    Using == on two Values does not compare the underlying values they represent,
    but rather the contents of the Value structs.
    To compare two Values, compare the results of the Interface method.


paulzhol:
nm, this won't work and cause a
`panic: runtime error: comparing uncomparable type func(string) string`.
Functions can only be compared to nil.

PR #342, opened 2017-07-22
introduce SetNameMapper to be used instead of assigning to NameMapper

paulzhol:
Document assigning to NameMapper still works only before sqlx is first used.
Remove the mprMu lock from mapper(), avoid comparing reflect.Value instances.
The result of such comparison is implementation specific.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"strings"
 	"sync"
+	"sync/atomic"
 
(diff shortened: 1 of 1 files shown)


jmoiron:
This is something I really want to do, and in general, I think the global things that sqlx exposed early on have all been frustrating limitations.  However, I'm reluctant to break compatibility :(

dlsniper:
This looks like a potential candidate for a new major version.

PR #348, opened 2017-08-25
Add Tx.NamedQueryContext method

acoshift:


yzzyx:
+1 for this to be merged. It would make switching between sqlx.DB and sqlx.Tx much easier, and the change is trivial and non-breaking.

dlsniper:
We are sorting the opened issues and pull requests, and I think this would be better served by #724.

PR #359, opened 2017-10-02
Expose Close method for single Row type

gomezjdaniel:
This is needed in order to support custom scanning implementations of the `sqlx.Row` type

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -228,6 +228,12 @@ func (r *Row) Columns() ([]string, error) {
 	return r.rows.Columns()
 }
 
+// Close allows to close the rows by exposing the Close method from the
+// underlying rows.
+func (r *Row) Close() error {
+	return r.rows.Close()
+}
+
 // Err returns the error encountered while scanning.
 func (r *Row) Err() error {
 	return r.err


gomezjdaniel:
cc @jmoiron

jmoiron:
The upstream library does not expose a Close, and `Row` is closed when you scan it.  Adding it might make it confusing, ie, it will lead people to believe that `Close` must be called, when actually it _must not_ be called after a scan.  I think I am against this unless you can provide a clear use case.

gomezjdaniel:
So as I said I need to close the underlying rows in the case where I am implementing a custom `sqlx.Row` scanner, in this case a scanner that supports to scan certain columns into an array/slice using a similar approach as the `pq` driver does with the [`Array`](https://godoc.org/github.com/lib/pq#Array) method.

As I stated before I need it for the situation where I want to implement a custom scanner for the `sqlx.Row` type.

My use case is that I have an `int32` slice that I want to write and scan into/from a single database column using a similar (almost equal) approach that the  [`Array`](https://godoc.org/github.com/lib/pq#Array) method from the `pq` library does.

Whereas in my package I have a similar method from the `func (*Row) scanAny(interface{}, bool) error` of this package that basically needs to place a `defer row.Close()` at the beginning for the case that `row.Scan(values....)` couldn't not be reached because of an error before of it.

Indeed this is what the sqlx.scanAny is [doing (L731)](https://github.com/jmoiron/sqlx/blob/master/sqlx.go#L731) but as sqlx.Row type has the underlying rows unexported there is no way I can close them without adding a `Close()` method for `sqlx.Row`.

PR #387, opened 2018-02-13
Added ColumnTypes support

cemremengu:
In addition to getting columns with `Columns` (already implemented), sql package also supports `ColumnTypes` which is handy (and I really need to use 😄 )


jmoiron:
Thanks!

cemremengu:
Thank YOU for the great work! 😄

Issue #390, opened 2018-02-24
How to register a bind type for a new driver?


cmoad:
In trying to integrate the "cloudsqlpostgres" driver, I don't see a way to specify the bind type as DOLLAR at setup time. I don't want to be forced to add code to manually rebind using sqlx.Rebind throughout my project which has hundreds of queries. Also, for some environments we use the "postgres" driver directly and I'm hoping to seamlessly switch between the two drivers.

Option 1: Add "cloudsqlpostgres" to the BindType function.
Option 2: Expose a way to manually set the bind type when setting up the connection.
Option 3: ???

Thanks for an awesome library.

ref: https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/proxy/dialers/postgres/hook_test.go
ref: https://github.com/jmoiron/sqlx/blob/master/bind.go#L22

cmoad:
Created #391 in case Option 1 is acceptable for the maintainers.

jmoiron:
An alternate to Option 1 would be nice at some point.  I think I'm going to wait for the versioning situation to settle before having backwards-incompatible fixes to some of the warts related to global registration et al, and this is one of them.

PR #391, opened 2018-02-24
added support for the cloudsqlpostgres driver

cmoad:
Google cloud has a driver for connecting to its postgres service using a secure proxy. Adding BindType support for this driver. This would address #390.

https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/proxy/dialers/postgres/hook_test.go


jmoiron:
Thank you very much!

PR #401, opened 2018-04-04
Added registering of bind types for custom drivers #400

evsan:


evsan:
also #390 and #391

krevels:
The inability to register a custom bind type is preventing me from using transaction-isolated drivers for my functional tests (such as [go-txdb](https://github.com/DATA-DOG/go-txdb)), where it requires that you register a new sql driver prior to use.  Unfortunately, this means that I'm unable to test code which uses named parameters at all.  It also requires me to explicitly rebind any queries using the `?` placeholder.

andreyvital:
@krevels my workaround for now is:
```Go
func init() {
	txdb.Register("pgx", "postgres", "postgres://postgres:postgres@localhost/postgres?sslmode=disable")
}

db := sqlx.MustOpen("pgx", "(...)")
```

Based off that: https://github.com/jmoiron/sqlx/blob/ee514944af4b0d1b90212831674e03df1b850998/bind.go#L26.

Not the best, but not the worst. Get you going. Hopefully a clean solution lands for solutions like txdb!

PR #420, opened 2018-06-07
Add support for sql.Conn (for go >= 1.9)

dmgk:


dmgk:
Doesn't currently builds with go1.8 because of Conn references in isUnsafe and mapperFor. I'm not sure how to better handle this - perhaps by moving these two functions to conditionally compiled (`+build !go1.9`) file?

dmgk:
Fixed by e2189dc, closing.

PR #421, opened 2018-06-13
Add sqlx.Connx to correspond to sql.Conn

henrylee2cn:
Requires version: go ≥ 1.9

### DIFF
--- a/sqlx_conn.go
+++ b/sqlx_conn.go
@@ -1,0 +1,137 @@
+package sqlx
+
+import (
+	"context"
+	"database/sql"
+)
+
+// Connx represents a single database connection rather than a pool of database
+// connections. Prefer running queries from DB unless there is a specific
+// need for a continuous single database connection.
+//
+// A Connx must call Close to return the connection to the database pool
+// and may do so concurrently with a running query.
+//
+// After a call to Close, all operations on the
+// connection fail with sql.ErrConnDone.
+type Connx struct {
+	*sql.Conn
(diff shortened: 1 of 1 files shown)


jmoiron:
Nice contribution.  I'd like to add some testing and perhaps get vgo integration going before I leave some verisons of Go behind.

henrylee2cn:
The test just like this:

```go
type Stat struct {
	Uri       string `db:"uri"`
	Total     int64  `db:"total"`
	CreatedAt int64  `db:"created_at"`
}

func TestDBConn(t *testing.T) {
	conn, err := db.Conn(context.Background())
	if err != nil {
		t.Fatal(err)
	}
	defer conn.Close()
	var total int64 = 1
	r, err := conn.ExecContext(
		context.Background(),
		`INSERT INTO stat VALUE ('/j',?,1528996897)
		ON DUPLICATE KEY UPDATE total=@total_tmp:=total+?;`,
		total, 1,
	)
	if err != nil {
		t.Fatal(err)
	}
	rowsAffected, err := r.RowsAffected()
	if err != nil {
		t.Fatal(err)
	}
	if rowsAffected == 2 {
		err = conn.GetContext(context.Background(), &total, "SELECT @total_tmp as total;")
		if err != nil {
			t.Fatal(err)
		}
	}
	t.Log(total)
}
```

glaslos:
@henrylee2cn how about adding your test to the PR?

PR #433, opened 2018-07-28
Copy MapScan caveat from doc page into function doc

renannprado:
Since the sqlx.MapScan(ColScanner, map[string]interface{}) error is usually not used by the users of this package, it's handy to have such documentation next to these functions so that the user can easily check the documentation from the method he/she is using.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -585,6 +585,9 @@ func (r *Rows) SliceScan() ([]interface{}, error) {
 }
 
 // MapScan using this Rows.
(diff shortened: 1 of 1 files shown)


renannprado:
@jmoiron do you know when this can be merged?

renannprado:
@jmoiron any idea if this will ever be merged?

renannprado:
@jmoiron could you please review it?

renannprado:
@jmoiron looks like you don't have interest in this contribution nor have the time to review it, I'll soon close it if there's no reply

PR #446, opened 2018-08-23
Add go.mod and go.sum for go module

flibustenet:
https://github.com/lib/pq/releases/tag/v1.0.0 was tagged v1.0.0 and with a `go.mod` file just now. I believe it's good time to add it to sqlx now.

I added the `go.mod` file with `go1.11rc1 mod init github.com/jmoiron/sqlx` then run `go1.11rc1 test all` it add all the dependencies and generate the `go.sum` with hashes of theses dependencies. I didn't test with `mysql` that i've not installed.

I think after that sqlx should be tagged with a semver compatible version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,8 @@
+module github.com/jmoiron/sqlx
+
+require (
+	github.com/go-sql-driver/mysql v1.4.0
+	github.com/lib/pq v1.0.0
+	github.com/mattn/go-sqlite3 v1.9.0
+	google.golang.org/appengine v1.1.0 // indirect
+)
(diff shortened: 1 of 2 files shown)


glaslos:
@jmoiron would be great to get this merged. Maybe even do a release afterwards?

jmoiron:
https://github.com/jmoiron/sqlx/releases/tag/v1.2.0

I'm not super sold on russ' "test deps should go along with normal deps".  Pre-modules, I'd tagged `sqlx-v1.0` and `sqlx-v1.1`, so in keeping with that I've tagged 1.2 which contains a number of fixes and sql server support.

Issue #447, opened 2018-09-03
Missing NamedQueryContext on *sqlx.Tx


samjohnduke:
Hi,

I'm not a prolific contributor, but I found this method to be missing on the transaction struct. I could just use a prepare query yes, but for consistency i think i should be there. I will create a quick pull request to fix. 

Sam

MichaelMallett:
Also curious about this one, has it been removed for a reason?

dackroyd:
There have been 3 attempts at PRs for this change:

https://github.com/jmoiron/sqlx/pull/348
https://github.com/jmoiron/sqlx/pull/373
https://github.com/jmoiron/sqlx/pull/448

This repository still seems to be active, with recent commits, so it's not clear what is holding this back.

Each of the above PRs lacks any unit tests, but those should be easily implemented?

JohnMaguire:
I'm also interested in this.

konart:
I'm pretty sure many people who actively use the package are interested.

@jmoiron is there anything that can be\should be done for this to be part of the package?

josnidhin:
I was creating an interface in my application so I reuse the same code for db interaction with and without transactions and that's when I realized `NamedQueryContext` is missing in `sqlx.Tx`. Any particular reason why this method is missing?

walles:
Six PRs have been made for this issue that I could find right now:

https://github.com/jmoiron/sqlx/search?q=NamedQueryContext+is%3Apr&type=issues

Also, this is the third highest voted issue out of 240 open ones:

https://github.com/jmoiron/sqlx/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

@jmoiron do you think you could do one of:
* If you can't / won't maintain this project any more:
  * Find one or more new maintainers. Maybe ask some of the biggest-contributors-except-yourself if they want in?
  * Or, put a note at the top of `README.md` that the project won't be maintained any more and archive it
* Look into one of the above PRs and have one of them merged

Either way, you built something useful, thanks!

JohnMaguire:
Just in case someone hasn't figured this out yet: it is possible to do `sqlx.NamedQueryContext(ctx, tx, query, arg)`. `tx` must impelement `sqlx.ExtContext`.

sakishrist:
`sqlx.NamedQueryContext(ctx, tx, query, arg)` with postgres resulted in question marks in the query instead of the postgres positional characters like `$1`, `$2`..

[...]

PR #448, opened 2018-09-03
update tx *Tx to include NamedQueryContext

samjohnduke:
As referenced in issue i created earlier as #447, this pull request ass the name query context method to the transaction struct.


Jammizzle:
Almost opened a PR doing this exact thing!

PR #453, opened 2018-09-23
adds support for structscan in multiple resultset queries

basvanbeek:
StructScan fails in a multiple result set scenario when the result sets have different row layouts.

Currently sqlx is unaware of iteration over multiple result sets and uses the first call to StructScan to set a field cache for performance reasons. When a next result set is iterated over, this field cache will most likely not match the struct provided.

By wrapping the NextResultSet method from sql.Rows and setting the field cache `started` property back to false the problem is solved.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -589,6 +589,16 @@ func (r *Rows) MapScan(dest map[string]interface{}) error {
 	return MapScan(r, dest)
 }
 
+// NextResultSet moves to the next resultset if available and resets the field cache.
+func (r *Rows) NextResultSet() bool {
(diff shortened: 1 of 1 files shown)


basvanbeek:
@jmoiron friendly ping. any feedback?

jpeirson:
@jmoiron is there a way forward on merging this PR? This fixes a critical issue for, what to me is, a common use case.

Dynom:
FYI; we've started using this branch @basvanbeek works great 👍

dlsniper:
We are sorting the opened issues and I think this can be closed in favor of #747.

PR #478, opened 2018-11-16
support scanAll *interface{}

fifsky:
We have a scenario where we call the parameter of the Select is an interface，For example:

```
type Users struct {
	Id        int       `db:"id"`
	Name      string    `db:"name"`
}

func main() {
	var m interface{}
	mm := make([]Users, 0)
	m = mm

	db, err := sqlx.Connect("mysql", "root:123456@tcp(127.0.0.1:3306)/blog?charset=utf8&parseTime=True&loc=Asia%2FShanghai")
	if err != nil {
		log.Fatalln(err)
	}

	err = db.Select(&m, "SELECT * FROM users")
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Printf("m===>%#v\n",m.([]Users))
}
```

But, Select first parameter must slice, We tried all sorts of ways to support this, and eventually we found that we could only make Elem calls to the Interface in the scanAll function, and then Scan of set Value

We have tested that the change is compatible and will not affect the slice parameter. We hope you can consider incorporating this feature or provide better Suggestions. Thank you！

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -900,7 +900,13 @@ func scanAll(rows rowsi, dest interface{}, structOnly bool) error {
 	if value.IsNil() {
 		return errors.New("nil pointer passed to StructScan destination")
 	}
-	direct := reflect.Indirect(value)
+	baseDirect := reflect.Indirect(value)
+	direct := baseDirect
+
+	if direct.Kind() == reflect.Interface {
+		value = direct.Elem()
+		direct = reflect.Indirect(value)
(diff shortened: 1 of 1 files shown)


jmoiron:
For posterity:

There isn't a need to pass `*interface{}`.  You can achieve this via:

```go
var m interface{}
mm := make([]User, ...)
m = &mm

...
db.Select(m, ...)
```

PR #501, opened 2019-03-20
add support for batch insert/exec with maps in addition to structs

J-Zeitler:
This is a feature specific use case I have, where I use the same internal logic for importing different tables into a db.

Thus I would like to use the syntax found in the test case added in [named_test.go](named_test.go)

Use case is something like this:

```go
type tableDump struct {
	tableName   string
	tableValues []map[string]interface{}
}

var dump tableDump
// ... read table data into `dump`

dbColKeys := safeParseKeys(dump)
placeholderKeys := prependEach(dbColKeys, ":")

_, err = db.NamedExec("INSERT INTO " + dump.tableName + "(" +
		strings.Join(dbColKeys, ",") + ")" +
		"VALUES " +
		"(" + strings.Join(placeholderKeys, ",") + ")",
                dump.tableValues,
)
```

### DIFF
--- a/named.go
+++ b/named.go
@@ -157,6 +157,10 @@ func bindAnyArgs(names []string, arg interface{}, m *reflectx.Mapper) ([]interfa
 // type, given a list of names to pull out of the struct.  Used by public
 // BindStruct interface.
 func bindArgs(names []string, arg interface{}, m *reflectx.Mapper) ([]interface{}, error) {
+	if maparg, ok := arg.(map[string]interface{}); ok {
+		return bindMapArgs(names, maparg)
+	}
+
 	arglist := make([]interface{}, 0, len(names))
 
 	// grab the indirected value of arg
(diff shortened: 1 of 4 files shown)


named.go
```diff
@@ -157,6 +157,10 @@ func bindAnyArgs(names []string, arg interface{}, m *reflectx.Mapper) ([]interfa
 // type, given a list of names to pull out of the struct.  Used by public
 // BindStruct interface.
 func bindArgs(names []string, arg interface{}, m *reflectx.Mapper) ([]interface{}, error) {
+	if maparg, ok := arg.(map[string]interface{}); ok {
```
arp242, line 160: `bindAnyArgs()` already has:

```
func bindAnyArgs(names []string, arg interface{}, m *reflectx.Mapper) ([]interface{}, error) {
    if maparg, ok := arg.(map[string]interface{}); ok {
        return bindMapArgs(names, maparg)
    }
    return bindArgs(names, arg, m)
}
```

jmoiron: There's only one use of bindArgs, and it seems safe to replace it with `bindAnyArgs`, so doing that in a separate branch.

PR #513, opened 2019-04-16
Prepared statement helper functions

sambengtson:
Background:  A few of my applications are very "prepared statement" heavy.  This streamlines some of the code that I was writing over and over again.  May help others.

Please feel free to change function names if you feel appropriate


sqlx.go
```diff
@@ -380,6 +380,51 @@ func (db *DB) PrepareNamed(query string) (*NamedStmt, error) {
 	return prepareNamed(db, query)
 }
 
+// Prepared statement helper.  Wraps Preparex
+func (db *DB) PrepareExecute(query string, args ...interface{}) (sql.Result, error) {
```
nussjustin, line 384: I don't see a point in these helpers.

Prepared statements are meant to be prepared once and reused, not to be prepared, executed and closed each time. Where is the added value compared to a simple call do `db.Exec(query, args...)`?

Same for the other helpers you added.

Please note that `db.Exec` and co. will automatically prepare a statement under the hood when you pass any arguments to the query¹. The prepared statement will be closed once the query finishes, which is exactly the same as what your helper does (except that it does not ignore the error when closing, which you should never do).

Basically your `PrepareExecute` can just be replaced with a call to `db.Exec(query, args...)`. Both will do basically the same (meaning that both will probably be extremely slow, since the query has to be prepared first and then closed afterwards), with `db.Exec` doing a better job at checking the errors.

The same is true for the other helpers.

Also your helpers are not taking a `context.Context`. Since the `context` package was a late addition to Go the `database/sql` package and other packages had to duplicate a lot of functions to support it, but all new functions should optimally be designed to take a `context.Context` from the beginning.

¹ There can be exceptions. I know that the [go-sql-driver/mysql](https://github.com/go-sql-driver/mysql) has a connection level option `interpolateParams` which when enabled will avoid preparing the statement when using `db.Exec[Context]` and `db.Query[Context]`, but that is optional and using (and reusing!) a prepared statement is still preferred.

sambengtson: @nussjustin No problem!  I appreciate your detailed feedback.

Issue #505, opened 2019-03-30
NamedExec batch INSERT does not work with ON CONFLICT


bretep:
Recently commited feature #285 does not support ON CONFLICT.

example:
```
sots := []*Sot{&sot1, &sot2}
db.NamedExec("INSERT INTO sot (sot_identifier, name)
			VALUES (:sot_identifier, :name)
			ON CONFLICT (sot_identifier) DO
				UPDATE SET (name) = (
				EXCLUDED.name)", sots)
```

Here is the query generated by `NamedExec`:
```
INSERT INTO sot (sot_identifier, name)
			VALUES ($1, $2)
			ON CONFLICT (sot_identifier) DO
				UPDATE SET (name) = (
				EXCLUDED.name),(
				EXCLUDED.name)
```

bretep:
@hmgle is this something you could fix easily?

bretep:
Looks like `sqlx.Named` also has the same issue.

hmgle:
This bug may be caused by the `fixBound` function. I will check it later.

hmgle:
This line [EndBracketsReg = regexp.MustCompile(`\([^()]*\)\s*$`)](https://github.com/jmoiron/sqlx/blob/cdf62fdf55f66fb2484fe214f89c059f0bd3f567/named.go#L214) assumes that the `value_list` part is at the end of `bound`. We cat fix the regexp like this:

```go
var valueBracketReg = regexp.MustCompile(`\([^(]*\?+[^)]*\)`)

func fixBound(bound string, loop int) string {
	loc := valueBracketReg.FindStringIndex(bound)
	if len(loc) != 2 {
		return bound
	}

	var buffer bytes.Buffer

	buffer.WriteString(bound[0:loc[1]])
	for i := 0; i < loop-1; i++ {
		buffer.WriteString(",")
		buffer.WriteString(bound[loc[0]:loc[1]])
	}
	buffer.WriteString(bound[loc[1]:])
	return buffer.String()
}
```

The patch:

```diff
-var (
-	EndBracketsReg = regexp.MustCompile(`\([^()]*\)\s*$`)
-)
+var valueBracketReg = regexp.MustCompile(`\([^(]*\?+[^)]*\)`)
 
 func fixBound(bound string, loop int) string {
-	endBrackets := EndBracketsReg.FindString(bound)
-	if endBrackets == "" {
+	loc := valueBracketReg.FindStringIndex(bound)
+	if len(loc) != 2 {
[...]

PR #514, opened 2019-04-23
Fix batch INSERT when using ON CONFLICT

bretep:
Fixes issue #505

### DIFF
--- a/named.go
+++ b/named.go
@@ -210,21 +210,21 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var (
-	EndBracketsReg = regexp.MustCompile(`\([^()]*\)\s*$`)
-)
+var valueBracketReg = regexp.MustCompile(`\([^(]*\?+[^)]*\)`)
 
 func fixBound(bound string, loop int) string {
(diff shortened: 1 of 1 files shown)


bretep:
The failure is unrelated to this commit and fails only for go1.8

bretep:
@hmgle are you able to merge this?

bretep:
I've been using this patch for the past 20 or so days. It's been working as expected. Thanks @hmgle

hmgle:
@jmoiron Could you please review and handle this pr? Thanks.

bretep:
Thanks @hmgle for pingin @jmoiron 

@jmoiron Hopefully we can get this committed soon so that I don't have to maintain a fork. :) Please let me know if I need to take any additional steps to get this PR accepted.

jmoiron:
The 1.8 failure is due to the mysql driver dropping support for 1.8;  I believe my backwards compatability guarantee also drops 1.8 so I'll remove it from the testing matrix instead of trying to fix it.

Additionally, since there has been no tagged release yet for this feature (I want to get more time to live with it), I don't mind removing an exported identifier...  in fact, I prefer it, it wasn't necessary to export in the first place.

bretep:
Thank you!

PR #520, opened 2019-06-02
Add function to specify bindtype for driver name

lamebear:
This allows for specifying the `bindtype` for a given `drivername` when that `drivername` does not already have a `bindtype` specified. This change allows for sqlx to be used with any driver, not just the ones hardcoded. If `BindDriver` is called with parameters that would change an existing binding, `ErrRebind` is thrown.

Example use case:
```
package main

import (
    github.com/jmoiron/sqlx
    _ github.com/lamebear/custom-db-driver
)

func main() {
    err := sqlx.BindDriver("mycustomdriver", sqlx.DOLLAR)
    if err != nil {
        log.Fatal(err)
    }
}
```

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -20,21 +20,39 @@ const (
 	AT
 )
 
+var (
+	ErrRebind = errors.New("binding already exists")
+
+	bindings = map[string]int{
+		"postgres":         DOLLAR,
+		"pgx":              DOLLAR,
+		"pq-timeouts":      DOLLAR,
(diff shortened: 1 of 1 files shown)


jmoiron:
Verison of this merged today, same interface but slightly different implementation, see https://github.com/jmoiron/sqlx/commit/00c6e74d816a1ce682afa7699a94c6b29b5f3c7c

PR #547, opened 2019-08-13
Handle error returned from a.Value()

### LINKED ISSUE #546: Surface errors returned from Valuer#Value in sqlx.In
Currently when iterating the args provided to `In`, any errors returned by `Value()` are ignored.

https://github.com/jmoiron/sqlx/blob/38398a30ed8516ffda617a04c822de09df8a3ec5/bind.go#L120

These errors should be surfaced in some way, as an error occuring during `Value` should fail binding.

Mungrel:
Fixes #546

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -117,7 +117,11 @@ func In(query string, args ...interface{}) (string, []interface{}, error) {
 
 	for i, arg := range args {
 		if a, ok := arg.(driver.Valuer); ok {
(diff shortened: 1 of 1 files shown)


Mungrel:
@jmoiron Thoughts?

Mungrel:
@jmoiron When will the next release be out?

PR #549, opened 2019-08-14
fix: null literals in JSON get unmarshaled as null string's bytes

danehammer:
JSONText's Unmarshal misintreprets a null literal as something to unmarshal.

I'm not 100% sure this is correct, definitely looking for feedback. But the root of why I feel like it's behaving wrong is this:

```go
type Foo struct {
  B []byte
  J types.JSONText
}

func main() {
        var f Foo
	fmt.Println("zero of B", f.B)
	fmt.Println("B is nil", f.B == nil)
	fmt.Println("zero of J", f.J)
	fmt.Println("J is nil", f.J == nil)	
	json.Unmarshal([]byte(`{"b": null, "j": null}`), &f)
	fmt.Println("after unmarshal B", f.B)
	fmt.Println("B is nil", f.B == nil)
	fmt.Println("after unmarshal J", f.J)
	fmt.Println("J is nil", f.J == nil)	
}
```

Which outputs:

```
zero of B []
B is nil true
zero of J []
J is nil true
after unmarshal B []
B is nil true
after unmarshal J [110 117 108 108]
J is nil false
```

The bytes `110 117 108 108` are the bytes of "null".

For ease of reproduction: https://play.golang.org/p/neMaVt39bRb

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -72,6 +72,11 @@ func (j *JSONText) UnmarshalJSON(data []byte) error {
 	if j == nil {
 		return errors.New("JSONText: UnmarshalJSON on nil pointer")
 	}
+
+	if string(data) == "null" {
+		return nil
+	}
+
 	*j = append((*j)[0:0], data...)
 	return nil
 }
--- a/types/types_test.go
+++ b/types/types_test.go
@@ -1,6 +1,9 @@
 package types
 
-import "testing"
+import (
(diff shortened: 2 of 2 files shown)


danehammer:
@jmoiron I'd say this is a beautiful PR, wouldn't you? 😄 27 lines

PR #551, opened 2019-08-16
Fix #473: handle nil arguments in In()

### LINKED ISSUE #473: Inserting a struct with sql.NullString returns panic
Hi there,
We've been using structs for returning fields with sql.NullString which has worked fine, however when using a struct that has an sql.NullString to insert into our database we get a panic:

```panic: reflect: call of reflect.Value.Type on zero Value```

It seems to be hitting the In function here:
```
v := reflect.ValueOf(arg)
t := reflectx.Deref(v.Type())
```

Thanks!

### PR DESCRIPTION
**Pull request by ojrac**
I ran in to #473, and figured out a fix. Please let me know if you'd like any changes to the diff.

Without the nil check in bind.go:102, the test panics with this stack:

```
panic: reflect: call of reflect.Value.Type on zero Value [recovered]
        panic: reflect: call of reflect.Value.Type on zero Value

goroutine 106 [running]:
testing.tRunner.func1(0xc00029e700)
        .../go/src/testing/testing.go:830 +0x392
panic(0x835440, 0xc00030cf20)
        .../go/src/runtime/panic.go:522 +0x1b5
reflect.Value.Type(0x0, 0x0, 0x0, 0xd8e240, 0x840460)
        .../go/src/reflect/value.go:1813 +0x169
github.com/jmoiron/sqlx.asSliceToExpand(0x0, 0x0, 0x0, 0x0, 0x0, 0x1)
        sqlx/bind.go:108 +0xc7
github.com/jmoiron/sqlx.In(0x8a4519, 0x2a, 0xc00030cea0, 0x2, 0x2, 0xc0000a05d0, 0x30, 0xc0000a7d80, 0x4, 0x4, ...)
        sqlx/bind.go:143 +0x127
github.com/jmoiron/sqlx.TestIn(0xc00029e700)
        sqlx/sqlx_test.go:1519 +0x4fa
testing.tRunner(0xc00029e700, 0x8ad6c0)
        .../go/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
        .../go/src/testing/testing.go:916 +0x35a
exit status 2
FAIL    github.com/jmoiron/sqlx 2.097s
```

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -98,6 +98,26 @@ func rebindBuff(bindType int, query string) string {
 	return rqb.String()
 }
 
+func asSliceToExpand(i interface{}) (v reflect.Value, ok bool) {
+	if i == nil {
+		ok = false
+		return
+	}
+
+	v = reflect.ValueOf(i)
+	t := reflectx.Deref(v.Type())
+
+	// []byte is a driver.Value type so it should not be expanded
+	ok = t.Kind() == reflect.Slice && t != reflect.TypeOf([]byte{})
+
+	if !ok {
+		// Don't return non-expandable values
+		v = reflect.Value{}
+	}
+
+	return
+}
+
 // In expands slice values in args, returning the modified query string
 // and a new arg list that can be executed by a database. The `query` should
 // use the `?` bindVar.  The return value uses the `?` bindVar.
@@ -119,11 +139,8 @@ func In(query string, args ...interface{}) (string, []interface{}, error) {
 		if a, ok := arg.(driver.Valuer); ok {
 			arg, _ = a.Value()
 		}
-		v := reflect.ValueOf(arg)
-		t := reflectx.Deref(v.Type())
 
-		// []byte is a driver.Value type so it should not be expanded
-		if t.Kind() == reflect.Slice && t != reflect.TypeOf([]byte{}) {
+		if v, ok := asSliceToExpand(arg); ok {
 			meta[i].length = v.Len()
 			meta[i].v = v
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by r3code** [comment]
Why it still not merged?

File: bind.go [hunk 1]
```diff
@@ -98,6 +98,26 @@ func rebindBuff(bindType int, query string) string {
 	return rqb.String()
 }
 
+func asSliceToExpand(i interface{}) (v reflect.Value, ok bool) {
+	if i == nil {
+		ok = false
```
cespare: This function is shorter and much easier to understand if it does not use naked returns.

```
func asSliceToExpand(i interface{}) (v reflect.Value, ok bool) {
	if i == nil {
		return nil, false
	}

	v = reflect.ValueOf(i)
	t := reflectx.Deref(v.Type())

	// []byte is a driver.Value type so it should not be expanded
	if t.Kind() != reflect.Slice || t == reflect.TypeOf([]byte{}) {
		return nil, false
	}
	return v, true
}
```

Also, a name like `asSliceForIn` might be clearer; "expand" is a bit vague.

cespare:
@jmoiron we ran into this bug when trying to upgrade sqlx.

I reviewed the code and it looks correct. (I have some style suggestions but they are minor.) There is also a test for the change which clearly demonstrates the problem.

If you could review and merge, that would be great!

bind.go
ojrac: Thanks for the naming idea -- `asSliceForIn` is much better. I updated the named returns' style as well, but overall made it a little longer by splitting the slice and `[]byte` checks into multiple lines; coming back to it after 3 months, it just felt too dense to understand what was going on.

Issue #559, opened 2019-09-04
Custom bind types for custom database drivers


mr337:
We are using sqlx through pop for our ORM. We needed to modify the pq driver to include a custom authentication with a small tweak and register as a new driver. We found that since the driver is registered and has a different name than `postgres` we are binded as `UNKOWN` causing issues for prepared queries.

There are a few issues such as https://github.com/jmoiron/sqlx/issues/400 and https://github.com/jmoiron/sqlx/pull/401 to register new binds. These look to be on hold by https://github.com/jmoiron/sqlx/issues/390#issuecomment-369124648.

Would anyone be interested in a PR to add options that are prefixed `customX` binding? Similar to https://github.com/jmoiron/sqlx/pull/391/files#diff-fbe43523c613db26c24df36d429e7ad3R24 following the pattern of `custompostgres`, `custommysql`, `customoracle`. This would allow a method for devs to use a custom driver without modifying sqlx directly.  The only downside is a custom driver would need to be registered as `customX`. 

Thoughts? Happy to put in a PR!

andy-kimball:
I'd also like to see this solved, as I've just implemented the open source [copyist](https://github.com/cockroachdb/copyist) SQL mocking library, and then ran into this issue when trying to use `sqlx` with it. I've put comments over in the #560 PR with another possible approach that would allow for a bit more naming flexibility than this proposal, but even this proposal is better than nothing at all.

jmoiron:
This fixed by `sqlx.BindDriver` merge from November.  It will be in 1.3.x

PR #560, opened 2019-09-04
Add custom bind types

mr337:
This PR adds custom bind types for each driver so one can easily register a new driver without sqlx code change. This is a crutch till https://github.com/jmoiron/sqlx/pull/520 is added when versioning and such is at a better point.

Related issue: https://github.com/jmoiron/sqlx/issues/559

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -23,15 +23,15 @@ const (
 // BindType returns the bindtype for a given database given a drivername.
 func BindType(driverName string) int {
 	switch driverName {
-	case "postgres", "pgx", "pq-timeouts", "cloudsqlpostgres", "ql":
+	case "postgres", "pgx", "pq-timeouts", "cloudsqlpostgres", "ql", "custompostgres":
 		return DOLLAR
-	case "mysql":
+	case "mysql", "custommysql":
 		return QUESTION
-	case "sqlite3":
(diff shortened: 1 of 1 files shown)


kcarlson:
I just added a simliar pr and found that this seems to work for my purposes:

mr337:
That looks great, unfortunately we use an ORM on top of sqlx and trying to **not** have fork the ORM too. Managing all the forks gets a bit much :D

andy-kimball:
I just ran into this problem as well, as part of implementing the open source [copyist](https://github.com/cockroachdb/copyist) SQL mocking library. These hardcoded binding values make it difficult to use a custom driver with `sqlx`.  @jmoiron, is there any approach to solving this issue that you're willing to accept into the `sqlx` codebase? Another idea besides the one in this issue is that if we fall into the "unknown" bucket, sqlx could do string contains operations to look for custom drivers. For example, any driver with the text "postgres" in its name would use PG semantics, any with "mysql" in its name would use MySQL semantics, etc. That'd be very simple and would unblock custom drivers, as long as they have naming flexibility.

jmoiron:
A way to add these at runtime has been added in November, `sqlx.BindDriver`

Issue #394, opened 2018-03-19
`compileNamedQuery` did not recognize comment


choichen:
I have a query like this:
```sql
# author(chen): some explanation
INSERT INTO table(foo, bar) VALUE (:foo, :bar)
```

and got a error when calling `ExecNamedContext`: 
`could not find name  in xxx`.

Seems `compileNamedQuery` did not recognize the beginning `#` and tries to parse `:`(followed by nothing) as a name, which go a empty name and then error occurred.

The working around is just don't use `:` from comments, but still i think this is a problem.

jhngrant:
Named query compilation doesn't parse SQL comments:
https://github.com/jmoiron/sqlx/blob/master/named.go#L239

Are you not able to move your comment from the SQL query string and place it inline in your Go code?

choichen:
@jhngrant thanks for suggestion. I don't like it either, but sadly we have many similar legacy queries in many different services. Comment in sql helps me fast-locating slow queries in business code.
i am considering using a db proxy to help, but that costs much time.

Vorlif:
I also stumbled across the problem. If you use Sqlite's full text search, the [colon is used to filter columns](https://www.sqlite.org/fts5.html#fts5_column_filters).

Therefore named queries do not work here either. Is there any way to escape the colon?

Issue #368, opened 2017-11-21
date format literal is treated as named sql parameter: sqlx.Named error: unexpected `:` while reading named param at 370


sunfmin:
![image](https://user-images.githubusercontent.com/1014/33067865-7dc71f94-ceea-11e7-9282-b224c14a84ef.png)

```
sqlx.Named error: unexpected `:` while reading named param at 370
```

bwmarrin:
I just ran into this myself.  Is there a work around for this problem?

bwmarrin:
So, looking in the code, 

https://github.com/jmoiron/sqlx/blob/master/named.go#L251

It looks like you can put in `::` as a work around.

Issue #350, opened 2017-09-07
Prepare named statement counts the same names as multiple parameters


omerkirk:
Hi, 

Named statement sqls permit map[string]interface{} as arguments however when parsing the statement and deciding the number of parameters the list contains duplicates which causes a database/sql error like the following: 

`sql: statement expects 1 inputs; got 6`

For dbs that accept positional parameters there is no issue however in named parameters this looks like a bug. Since the db parses the statement and returns the number without the duplicates. 

I can prepare a pull request that fixes the issue for NAMED parameters if you like.

Issue #304, opened 2017-04-07
Add the possibility to chose a custom prefix for named parameters


skoot:
We use postgres and we have a bunch of queries in which we use the ::TEXT or ::INT syntax. Escaping them isn't convenient, and we just wouldn't have to if we were able to chose another character for the named params (like "@"). 

Letting users chose their own character prefix for named variable would be neat.

It could very easily be done on the package level by using a global variable, or it could be set on the connection level and then passed all the way down to the compileNamedQuery function.

What do you think?

jhngrant:
@skoot If you're using the pq driver ( github.com/lib/pq ), it uses the Postgres-native ordinal markers:

```
rows, err := db.Query(`SELECT name FROM users WHERE favorite_fruit = $1
	OR age BETWEEN $2 AND $2 + 3`, "orange", 64)
```
```
err := db.QueryRow(`SELECT $1::uuid`, s).Scan(&scanned)
```

skoot:
I'm using pgx and I use sqlx precisely to avoid ordinal markers and use named variables instead.

jhngrant:
@skoot My apologies. I misunderstood your original issue.

I'd agree, the historical Postgres type cast syntax is keyboard friendly when typing SQL statments:
```
SELECT 1234::TEXT
```
However, to keep the sqlx library lean there may be a counter argument of switching to standard SQL for production code:
```
SELECT CAST(1234 AS TEXT)
```
This would of course avoid the parsing issue you are seeing with prepared statements when using the ```::``` type cast syntax.

https://www.postgresql.org/docs/current/static/sql-expressions.html#SQL-SYNTAX-TYPE-CASTS

sachinwalia2k8:
@skoot can you please share how are you bridging sqlx with pgx. I am trying to find a working example where I could use two features from sqlx with pgx namely - struct scan and named parameters.

skoot:
Sure, it's an easy 2 step process: You start by getting a database/sql object from a pgx connection or pool, as explained here:

https://godoc.org/github.com/jackc/pgx/stdlib

Then you wrap it up in sqlx as you would any other database/sql object.

PR #579, opened 2019-12-05
Improve parsing of named parameters

joeandaverde:
To ignore comments, understand Unicode, and have fewer false positives.

This doesn't meet the original contract at the moment. The original expected any colon to be doubled up to be escaped. The double-colon (`::` -> `:`) wasn't strictly necessary but seemed to be as a result of not properly parsing string literals and comments.

It's probably worth maintaining backward compatibility but having to double-colon could trip up someone new to the package.

There's also likely more cases of how string literals can be defined based on the driver. This implementation understands Postgres.

Related issues:
https://github.com/jmoiron/sqlx/issues/394
https://github.com/jmoiron/sqlx/issues/368
https://github.com/jmoiron/sqlx/issues/304
https://github.com/jmoiron/sqlx/issues/302
https://github.com/jmoiron/sqlx/issues/206
https://github.com/jmoiron/sqlx/issues/166

This one can be done fairly easily too:
https://github.com/jmoiron/sqlx/issues/350

### DIFF
--- a/named.go
+++ b/named.go
@@ -14,12 +14,12 @@ package sqlx
 import (
 	"bytes"
 	"database/sql"
-	"errors"
 	"fmt"
 	"reflect"
 	"regexp"
 	"strconv"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/jmoiron/sqlx/reflectx"
 )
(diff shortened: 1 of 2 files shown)


joeandaverde:
@jmoiron Thoughts on this PR? Do you think it's a worthwhile change?

joeandaverde:
Hi, @jmoiron is there any chance this PR will be considered? Is there any more I can do to assist with closure?

PR #591, opened 2020-02-04
feat: upgrade mysql db driver and other libraries

akkumar:
Upgrade database mysql driver. 

Bunch of other package upgrades as well.


jmoiron:
updated these separately;  stayed on go-sqlite v1.x line

PR #565, opened 2019-10-07
[issue-447] added NamedQueryContext method on sqlx.Tx

meh-:

PR #617, opened 2020-05-12
Add missing NamedQueryContext function on slqx.Tx

taywrobel:
Adds the NamedQueryContext function to the Tx object. The function
already exists on the Db object, and the supporting code for
transactions is already in place, but this function is missing from
the set of named sqlx operations.


taywrobel:
Missed that this was already in PR in https://github.com/jmoiron/sqlx/pull/565

Closing this in favor of that.  Motion there would be much appreciated!

Issue #623, opened 2020-05-26
`NamedQueryContext` does not support `IN` clause with array


jrosspaperless:
Assume a simple query:

`query := SELECT * FROM places WHERE id IN (:placeIDs)`

where `placeIDs` is `[]uuid.UUID` and queryArgs is `map[string]interface{}`

To execute this one would need to:

```
func executeInClauseSupport(ctx context.Context, db *sqlx.DB, query string, queryArgs map[string]interface{}) (*sql.Rows, error) {
	q, args, err := sqlx.Named(query, queryArgs)
	if err != nil {
		return nil, err
	}
	q, args, err = sqlx.In(q, args...)
	if err != nil {
		return nil, err
	}
	q = db.Rebind(q)
	return db.QueryContext(ctx, q, args...)
}
```

because an error is thrown when trying to execute:

`rowsCount, errCount := d.db.NamedQueryContext(ctx, queryCount.String(), queryArgs)`

jmoiron:
I commented on #623, agreed it fixes this but I feel like it causes too many other issues.  I feel like `sqlx.In`, `sqlx.Rebind` and the Named query stuff could all be implemented in a cleaner way, but not without sacrificing backwards incompatibility.  As they are, they are built organically on top of each other over the course of several years;  I think the answer here is to create a helper function for your desired combination of `In` and `NamedQuery`;  I don't like the idea of `In` happening without people knowing that it's happening (or people "paying" for it unknowingly).

PR #624, opened 2020-05-26
A quick and dirty solution For Issue #623

jrosspaperless:
This simply runs the context through a thing that checks if the underlying type is an array.

https://github.com/jmoiron/sqlx/issues/623


jmoiron:
Having `NamedQuery` doing `In` under the covers feels unexpected;  no other Query functions do this.  It looks like this could be a convenience function in user code;  I think that's the only way to do this for now.

PR #632, opened 2020-06-13
Small performance improvement for batch insert

GSokol:
One time memory allocation is better than in loop


jmoiron:
Looks great, thanks!

PR #643, opened 2020-08-18
feat: support log

yongPhone:
Change-Id: Ic1388cd6470706ab4c4a0c06259ec7370b5c83e7


arp242:
I don't think this should depend directly on logrus.

This has been proposed before, and generally people have said this is not really something that needs to be in sqlx as you can use your own sqlx.DB wrapper if you need it.

jmoiron:
Agreed with the prior comment.  `database/sql` does not even use the default logging package.  I'm not sure it's appropriate to use a specific one, one of the goals with `sqlx` is that it is appropriate to use in any codebase that is using `database/sql`.

PR #660, opened 2020-10-20
Expose ValueBracketReg

### LINKED ISSUE #659: Cannot replace EndBracketsReg
We use postgist query like this: `INSERT INTO data (map_id, coordinates) VALUES (:map_id, GeomFromEWKB(:coordinates))`
It was working before with this hack:
```go
originalRegex := sqlx.EndBracketsReg
sqlx.EndBracketsReg = regexp.MustCompile(`\([^()]*\)\s*$|\([^)]*(GeomFromEWKB\(\?\))+.*\)\s*$`)
defer func() {
   sqlx.EndBracketsReg = originalRegex
}()
```
Since PR #514, EndBacketsReg was replaced by valueBracketReg, but this variable is not public.

main.go
```go
package main

import (
	"fmt"
	"log"

	"github.com/cridenour/go-postgis"
	"github.com/jmoiron/sqlx"
	_ "github.com/lib/pq"
)

type data struct {
	DataID int64          `db:"data_id"`
	MapID  int64          `db:"map_id"`
	Coord  postgis.PointS `db:"coordinates"`
}

func main() {
	db := sqlx.MustOpen("postgres", fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=disable",
		"localhost", 5432, "user", "pass123", "db"))

	dd := []data{
		{MapID: 1, Coord: postgis.PointS{SRID: 4326, X: -117.16686667, Y: 36.42191667}},
		{MapID: 2, Coord: postgis.PointS{SRID: 4326, X: -120.16686667, Y: 37.42191667}}}

	query := `
	INSERT INTO data (map_id, coordinates)
	VALUES (:map_id, GeomFromEWKB(:coordinates))`

	_, err := db.NamedExec(query, dd)
	log.Println(err)
}
```

docker-compose.yaml
```yaml
version: "3.8"

services:
  db:
    image: postgis/postgis:10-3.0
    environment:
      POSTGRES_DB: db
      POSTGRES_USER: user
      POSTGRES_PASSWORD: pass123
      POSTGRES_MULTIPLE_EXTENSIONS: postgis,hstore,postgis_topology,postgis_raster,pgrouting
    ports:
      - "5432:5432"
```

data.sql
```sql
CREATE TABLE data" (
	data_id bigserial NOT NULL,
	map_id int8 NOT NULL,
	coordinates geography(POINT,4326) NULL
)
```

go.mod
```
github.com/cridenour/go-postgis v1.0.0
github.com/jmoiron/sqlx v1.2.1-0.20200615141059-0794cb1f47ee
github.com/lib/pq v1.2.0
```

hexadecy:
Fix #659

### DIFF
--- a/named.go
+++ b/named.go
@@ -210,10 +210,10 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valueBracketReg = regexp.MustCompile(`\([^(]*\?+[^)]*\)`)
+var ValueBracketReg = regexp.MustCompile(`\([^(]*\?+[^)]*\)`)
 
 func fixBound(bound string, loop int) string {
-	loc := valueBracketReg.FindStringIndex(bound)
+	loc := ValueBracketReg.FindStringIndex(bound)
 	if len(loc) != 2 {
 		return bound
 	}


arp242:
I'm not sure if I follow what this fixes exactly; your issue doesn't really mention what the problem is? It would be helpful if you could expand on that.

I also wonder if this is something that can't be fixed better instead of letting people override some mysterious regexp? Your solution is also racy and people *will* run in to problems if they use sqlx in more than on goroutine.

hexadecy:
With the current regexp 
```go
var valueBracketReg = regexp.MustCompile(`\([^(]*\?+[^)]*\)`)
```
and with this sql request:
```
INSERT INTO data (map_id, coordinates) VALUES (:map_id, GeomFromEWKB(:coordinates))
```
we have this error:
```bash
pq: syntax error at end of input
```

Another possible fix is to have a more permissive regexp for any functions not just GeomFromEWKB(:coordinates), but I not sure of the security issue (sql injection). So I think we can at least go back to the previous behavior, right now we are stuck using an older version.

hexadecy:
@arp242 I updated the issue to give more info.

hexadecy:
Fix in #734

Issue #648, opened 2020-09-11
Bulk insertion of NamedExecContext does not support the NOW function.


wencan:
code:
```go
	dts := []struct {
		ID int64 `db:"id"`
	}{
		{
			ID: 20,
		},
		{
			ID: 21,
		},
	}
	_, err = dbx.NamedExecContext(context.Background(), `INSERT INTO time_zone (id, dt) VALUES(:id, NOW())`, dts)
```

got error:
```
Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '' at line 1
```

Issue #640, opened 2020-08-03
Bulk insert generate wrong SQL statement


dmngu9:
This is my insert query 
```
INSERT INTO
			user_books(
				user_id,
				status,
				books,
				updated_ts
			)
		VALUES (:user_id, :status, :books, NOW())
		ON CONFLICT(user_id)
		DO UPDATE SET
			books = EXCLUDED.books,
			updated_ts = NOW()
		WHERE
			user_books.user_id = EXCLUDED.user_id
```

This is the error `pq: syntax error at or near \"ON\"`

Query generated:

```
INSERT INTO
			user_books(
				user_id,
				status,
				books,
				updated_ts
			)
		VALUES ($1, $2, $3, NOW(),($4, $5, $6, NOW(),($7, $8, $9, NOW())
		ON CONFLICT(user_id)
		DO UPDATE SET
			books = EXCLUDED.books,
			updated_ts = NOW()
		WHERE
			user_books.user_id = EXCLUDED.user_id
```

theres missing bracket between each value to insert

Issue #622, opened 2020-05-26
Batch insert cannot evaluate SQL expression


dqn:
Single record insert can use SQL expressions in the VALUES clause, but batch insert added in #285 cannot.

e.g.

```go
user := &User{Name: "foo"}
db.NamedExec("INSERT INTO users (name, created_at, updated_at) VALUES (:name, NOW(), NOW())", user) // ok

users := []*User{
	{Name: "bar"},
	{Name: "baz"},
}
db.NamedExec("INSERT INTO users (name, created_at, updated_at) VALUES (:name, NOW(), NOW())", users) // pq: syntax error at end of input
```

rcornacchia:
I'm also running into an issue with bulk insert but inserting one row works.

dmngu9:
can you post the generated query from it?

dqn:
@dmngu9 This.

```go
user := &User{Name: "foo"}
s1, _, _ := db.BindNamed("INSERT INTO users (name, created_at, updated_at) VALUES (:name, NOW(), NOW())", user)
fmt.Println(s1) // => INSERT INTO users (name, created_at, updated_at) VALUES ($1, NOW(), NOW())

users := []*User{
	{Name: "bar"},
	{Name: "baz"},
}
s2, _, _ := db.BindNamed("INSERT INTO users (name, created_at, updated_at) VALUES (:name, NOW(), NOW())", users)
fmt.Println(s2) // => INSERT INTO users (name, created_at, updated_at) VALUES ($1, NOW(),($2, NOW(), NOW())
```

dmngu9:
same issue as mine. Theres sth wrong with query with built in function . https://github.com/jmoiron/sqlx/issues/640

dmngu9:
if you replace NOW() with pq.FormatTime(time.Now()), it would work. Something wrong with NOw() affect generate query string

carmel:
I encountered a similar problem:
```golang
test := Test{
  Name: " John",
  Age:  30,
  Id:   "bskdvfjreo018g2c5pqg",
}
// table test has fields: id, name, age
_, err := sqlxdb.DB.NamedExec("NSERT INTO test(name,age,id)VALUES(:Name,:Age,:Id)", &test)
log.Println(err)

// log as follow:

// 2020/08/07 11:49:15 could not find name Name in &sqlx.Test{Id:"bskdvfjreo018g2c5pqg", Name:"Vector", CreateAt:0, Age:30}
```

[...]

PR #667, opened 2020-11-18
Fix RegEx for multi-line inserts

romandvoskin:
Fixes issues with multi line inserts. #648, #640, #622

### DIFF
--- a/named.go
+++ b/named.go
@@ -210,7 +210,7 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valueBracketReg = regexp.MustCompile(`\([^(]*\?+[^)]*\)`)
+var valueBracketReg = regexp.MustCompile(`\([^(]*.[^(]\)$`)
 
 func fixBound(bound string, loop int) string {
 	loc := valueBracketReg.FindStringIndex(bound)
--- a/named_context_test.go
+++ b/named_context_test.go
@@ -9,7 +9,7 @@ import (
 )
 
 func TestNamedContextQueries(t *testing.T) {
-	RunWithSchema(defaultSchema, t, func(db *DB, t *testing.T) {
+	RunWithSchema(defaultSchema, t, func(db *DB, t *testing.T, now string) {
 		loadDefaultFixture(db, t)
 		test := Test{t}
 		var ns *NamedStmt
--- a/named_test.go
+++ b/named_test.go
@@ -2,6 +2,7 @@ package sqlx
 
 import (
 	"database/sql"
+	"fmt"
 	"testing"
 )
 
@@ -119,7 +120,7 @@ func (t Test) Errorf(err error, format string, args ...interface{}) {
 }
 
 func TestNamedQueries(t *testing.T) {
-	RunWithSchema(defaultSchema, t, func(db *DB, t *testing.T) {
+	RunWithSchema(defaultSchema, t, func(db *DB, t *testing.T, now string) {
 		loadDefaultFixture(db, t)
 		test := Test{t}
 		var ns *NamedStmt
@@ -191,8 +192,8 @@ func TestNamedQueries(t *testing.T) {
 			{FirstName: "Ngani", LastName: "Laumape", Email: "nlaumape@ab.co.nz"},
 		}
 
-		_, err = db.NamedExec(`INSERT INTO person (first_name, last_name, email)
-			VALUES (:first_name, :last_name, :email)`, sls)
+		insert := fmt.Sprintf("INSERT INTO person (first_name, last_name, email, added_at) VALUES (:first_name, :last_name, :email, %v)", now)
+		_, err = db.NamedExec(insert, sls)
 		test.Error(err)
 
 		for _, p := range sls {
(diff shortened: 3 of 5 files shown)


romandvoskin:
@jmoiron Please take a look. The fix is in the RegEx. `now` is added to test harness to encourage sql function testing.

jmoiron:
fixes #648, #640, #622

aarengee:
@jmoiron  This regex assumes the value_list to be inserted is at the end of the list and has reopened a previous issue #505

aarengee:
For anyone looking for a solution I was able to make use of inner functions and make a bulk upsert work

```go
	query, queryArgs, _ := db.BindNamed(queryWithoutONCONFLICTWhichEndsInBracket, driverPreferences)
	query = db.Rebind(queryWithoutONCONFLICTWhichEndsInBracket)
	query = queryWithoutONCONFLICTWhichEndsInBracket + onConflictStatement
	rows, _ := repo.db.Queryx(query, queryArgs...)
```
This should work . 

cc : @matthisk

smoussa:
@aarengee Unfortunately that doesn't work for me. I get `pq: syntax error at or near ":"`. Here is the only place I use the colon syntax:

```
... VALUES (:col1, :col2, ...) ... ON CONFLICT
```

aarengee:
Hey @smoussa Let me give a **Complete Example**

```golang
	sliceOfStructs := []model.DriverPreference{
		{
		UserID:   "123",
		Name: "aarengee",
		Address: "xd",
		Eligible: true,
		},
		{
			UserID:   "1234",
			Name: "aarengeeAgain",
			Address: "xd",
			Eligible: false,
		},
	}
	upsertQuery := "INSERT INTO driver_preferences (user_id, name, address, eligible,updated_at) VALUES (:user_id,:name,:address,:eligible,now())"
	onConflictStatement := " ON CONFLICT (user_id, name) DO UPDATE SET address = excluded.address,eligible = excluded.eligible, updated_at = now() RETURNING *"
	query, queryArgs, _ := repo.db.BindNamed(upsertQuery, sliceOfStructs)
	query = repo.db.Rebind(query)
	query = query + onConflictStatement
	rows, err := repo.db.Queryx(query, queryArgs...)
```
As stated I am using potsgres and sqlx@1.3.0 with pq@1.2.0. Although DB flavor and pq shouldnt matter. Hope this helps.
Maybe you can assist with a somewhat working code and we can work.

smoussa:
@aarengee Thanks for this. You'll notice that you've used the bound `query` string in `db.Rebind(query)`, whereas in your first example you've used `db.Rebind(queryWithoutONCONFLICTWhichEndsInBracket)`.

Your example has fixed the colon issue, but I've run into another problem when upserting two struct objects:
```
pq: got 16 parameters but the statement requires 8
```

This is the case when bulk upserting more than one item. In your example with two struct objects you would get:
```
pq: got 8 parameters but the statement requires 4
```

smoussa:
Using @abraithwaite's PR https://github.com/jmoiron/sqlx/pull/718, I've managed to find a solution. The implementation is also simpler:

```golang
query := "INSERT INTO ... VALUES ... ON CONFLICT ... DO UPDATE SET ... RETURNING ..."
rows, err := db.NamedQuery(query, sliceOfStructs)
```

Thanks @aarengee nonetheless! Hope this helps.

aarengee:
> This is the case when bulk upserting more than one item. In your example with two struct objects you would get:
> 
> ```
> pq: got 8 parameters but the statement requires 4
> ```
Hey @smoussa , If you see in my example I am indeed bulk upserting with  two structs. I was able to succesfully do the bulk upsert with two structs in my codebase.

P.S. I just deployed this code to PROD like 2 hours ago and its working just fine. As you know I can only use stable releases in a production ready code base . 

> Using @abraithwaite's PR #718, I've managed to find a solution. The implementation is also simpler:

Yes, This is indeed the right way and how @jmoiron intended it to be used. This was fixed then broken and now has been fixed again. 🤞 .

Hopefully this makes its way into a stable version release and I can use the same instead of the hack I managed to pull together.

venkatsafetyculture:
@aarengee tried your solution and still get `pq: ON CONFLICT DO UPDATE command cannot affect row a second time`, I have the following query 
`	sliceOfStructs := []model.Gamer{
		{
		UserID:   "123",
		Name: "aarengee",
		Address: "xda",
		Eligible: true,
		},
		{
			UserID:   "123",
			Name: "aarengee",
			Address: "xdb",
			Eligible: false,
		},
	}
	upsertQuery := "INSERT INTO gamer_details (user_id, name, address, eligible,updated_at) VALUES (:user_id,:name,:address,:eligible,now())"
	onConflictStatement := " ON CONFLICT (user_id, name) DO UPDATE SET address = excluded.address,eligible = excluded.eligible, updated_at = now() RETURNING *"
	query, queryArgs, _ := db.BindNamed(upsertQuery, sliceOfStructs)
	query = db.Rebind(query)
	query = query + onConflictStatement
	rows, err := db.Queryx(query, queryArgs...)`
I was hoping that i will have 1 record for user id 123 with address = xdb and eligible as false

squillace91:
I kept still getting the error when doing a batch insert with ON CONFLICT, so I found a solution that worked for me with a help of a friend. This was very hard to find but you can see more info here. https://www.prisma.io/dataguide/postgresql/inserting-and-modifying-data/insert-on-conflict#using-the-do-update-action

`DO UPDATE SET (var1, var2, var3, var4) = (excluded.var1, excluded.var2, excluded.var3, excluded.var4)`

Hope this helps!

PR #683, opened 2021-01-14
help fixBound identify "((?),(?))" and ("VALUES (?, CAST (? AS timest…

terryh:
…amp));"


terryh:
hope fixBound to identify 
"((?),(?))"
 "VALUES (?, CAST (? AS timestamp));"

Issue #713, opened 2021-03-17
NamedExec unable to bulk insert with functions


Mungrel:
Example query:

```sql
INSERT INTO some_table (
 name,
 value
) VALUES (
 :name,
 handle_value(:value)
)
```

It looks to be due to [the regex](https://github.com/jmoiron/sqlx/blob/master/named.go#L227) matching the opening bracket of the `handle_value` function.

Mungrel:
I don't think there's any way to solve this effectively with Regex, and instead the parentheses should be parsed

ddouglas:
Yeah, we have this same issue when using `CURDATE()` & `NOW()`

yyh-gl:
`UUID_TO_BIN()` too 📝

jmoiron:
fixed in v1.3.2

Issue #712, opened 2021-03-17
NamedExec unable to form bulk insert query


RajatBanerjee:
in v 1.2.1 while performing a NamedExec with a slice of interface

`INSERT INTO foo (foo_nm,  ins_user) VALUES (:foo_nm, :ins_user)
ON DUPLICATE KEY UPDATE 
foo_nm = values(foo_nm),upd_user = values(ins_user)`

data := []interface{}`
`_, err := sqlx.NamedExec(db, query, data)`

would produce via `bindNamedMapper` (assume data had 2 values)
`INSERT INTO foo (foo_nm,  ins_user) VALUES (?, ?).(?,?)
ON DUPLICATE KEY UPDATE 
foo_nm = values(foo_nm),upd_user = values(ins_user)`

However in 1.3.1
This becomes 
`INSERT INTO foo (foo_nm,  ins_user) VALUES (?, ?)
ON DUPLICATE KEY UPDATE 
foo_nm = values(foo_nm),upd_user = values(ins_user),(ins_user)`

Notice the extra `ins_user` instead of (?,?) after values

I suspect its due to
https://github.com/jmoiron/sqlx/blob/master/named.go#L227
which was `regexp.MustCompile(`\([^(]*\?+[^)]*\)`)`

Mungrel:
That regex also causes issues when trying to use SQL procedures/functions in a named query https://github.com/jmoiron/sqlx/issues/713

RajatBanerjee:
Reverting back to 1.2.0 
However in 1.2.0 there is an issue with []map[string]interface{} , had to use a struct with `db` fields

jmoiron:
should be fixed with v1.3.2

Issue #709, opened 2021-03-13
Bulk insert with only one colum, invalid SQL is used


podhmo:
Bulk insert query with only one column,  generated code is something wrong. 

For example, `Named()` function is using like following.

```go
sqlx.Named("INSERT INTO vs (k) VALUES (:k)", []V{{K: "x"}, {K: "y"}})
```

Expected  SQL is here.

```
INSERT INTO vs (k) VALUES (?),(?)
```

But, generated SQL is something like this.

```
INSERT INTO vs (k) VALUES (?),(k) VALUES (?)
```

## Minimum reproducible code is here.

```go
package main

import (
	"log"

	"github.com/jmoiron/sqlx"
)

func main() {
	type V struct {
		K string `db:"k"`
		V int    `db:"v"`
	}

	{
		s, values, err := sqlx.Named("INSERT INTO vs (k) VALUES (:k)", []V{{K: "x"}, {K: "y"}})
		log.Printf("\n\tquery=%q\n\tvalues=%#+v\n\terr=%+v", s, values, err)
	}
	{
		s, values, err := sqlx.Named("INSERT INTO vs (k, v) VALUES (:k, :v)", []V{{K: "x"}, {K: "y"}})
		log.Printf("\n\tquery=%q\n\tvalues=%#+v\n\terr=%+v", s, values, err)
	}
}
``` 

the results.

```
2021/03/13 18:35:37 
	query="INSERT INTO vs (k) VALUES (?),(k) VALUES (?)"
	values=[]interface {}{"x", "y"}
	err=<nil>
2021/03/13 18:35:37 
	query="INSERT INTO vs (k, v) VALUES (?, ?),(?, ?)"
	values=[]interface {}{"x", 0, "y", 0}
	err=<nil>
```

podhmo:
https://github.com/jmoiron/sqlx/blob/a1d5e64734233358bc4e0a54beddc18509071a95/named.go#L227
This regular expression does not support 1-column input.

podhmo:
related to #694

podhmo:
Ah, OK, the regular expression ending with`$` is also bad.

podhmo:
https://github.com/jmoiron/sqlx/pull/667#issuecomment-787731678

podhmo:
Hmm, we need to clarify the spec.
If my understanding is correct, supporting grammar is something like below  (correct me if I'm wrong).

```diff
diff --git a/named.go b/named.go
[...]

Issue #705, opened 2021-02-24
Bulk insert using NamedExec breaks after upgrading from 1.2.0 to 1.3.1


davechallis:
The following worked in 1.2.0:

```
type Foo struct {
    A uint32
    B uint32
}

foos := []Foo {
    {A: 1, B: 1},
    {A: 2, B: 2},
    {A: 3, B: 3},
    {A: 4, B: 4},
    {A: 5, B: 5},
}

if _, err := db.NamedExec(`INSERT INTO x (a, b) VALUES (:a, :b)`, foos); err != nil {
    panic(err)
}
```
But after upgrading to 1.3.1 gives the error:

```
expected 2 arguments, got 5
```

aarengee:
https://github.com/jmoiron/sqlx/pull/667#issuecomment-787731678

jmoiron:
Should work in v1.3.2

PR #718, opened 2021-03-31
NamedExec Bulk Insert Fix

### LINKED ISSUE #657: NamedExec/slice does not work with Postgres INSERT ... ON CONFLICT
First of all, thanks for this fantastic lib, Jason! It has saved me countless hours in implementing database interfaces!

I recently came across a new(?) feature which allows a NamedExec to receive a slice of structs. This works beautifully with a simple INSERT statement, generating a `VALUES (...), (...), (....) ...` list.

However, it does not work with the Postgres `INSERT INTO table VALUES (...) ON CONFLICT DO UPDATE SET column1 = :gofield1, column2 = :gofield2, ...` type of statement, as the arguments for the latter part are not "multiplied out" for all slice elements, but only appended once to the args array.

abraithwaite:
This change-set is how I addressed the issues of bulk inserts not being handled cleanly when they're not at the end of the query.

It works by first adding `VALUES` to the expression (not sure if this is universally applicable to all supported SQL dialects) then adding a submatch to the regular expression for the fields we want to repeat.  Using the submatch, we're able to repeat the fields that we're concerned with to expand the insertion parameters.

Likely? Fixes #657, #694, #705, #709, #712, #713 ?

### DIFF
--- a/named.go
+++ b/named.go
@@ -224,21 +224,28 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valueBracketReg = regexp.MustCompile(`\([^(]*.[^(]\)\s*$`)
+var valueBracketReg = regexp.MustCompile(`VALUES\s+(\([^(]*.[^(]\))`)
 
 func fixBound(bound string, loop int) string {
-	loc := valueBracketReg.FindStringIndex(bound)
-	if len(loc) != 2 {
+	loc := valueBracketReg.FindAllStringSubmatchIndex(bound, -1)
+	// Either no VALUES () found or more than one found??
+	if len(loc) != 1 {
+		return bound
+	}
+	// defensive guard. loc should be len 4 representing the starting and
+	// ending index for the whole regex match and the starting + ending
+	// index for the single inside group
+	if len(loc[0]) != 4 {
 		return bound
 	}
 	var buffer bytes.Buffer
 
-	buffer.WriteString(bound[0:loc[1]])
+	buffer.WriteString(bound[0:loc[0][1]])
 	for i := 0; i < loop-1; i++ {
 		buffer.WriteString(",")
-		buffer.WriteString(bound[loc[0]:loc[1]])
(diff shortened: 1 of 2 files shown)


named.go
```diff
@@ -224,21 +224,28 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
...
-	if len(loc) != 2 {
+	loc := valueBracketReg.FindAllStringSubmatchIndex(bound, -1)
+	// Either no VALUES () found or more than one found??
+	if len(loc) != 1 {
+		return bound
+	}
+	// defensive guard. loc should be len 4 representing the starting and
+	// ending index for the whole regex match and the starting + ending
+	// index for the single inside group
+	if len(loc[0]) != 4 {
```
abraithwaite, line 238: If `loc[0] != 4`, that indicates that the regex was changed and either the submatcher was removed or another submatcher was added.

gsora:
Just tried this patch myself and it works fine with `INSERT ... ON CONFLICT DO UPDATE SET`.

named_test.go
```diff
@@ -214,7 +218,7 @@ func TestNamedQueries(t *testing.T) {
 		}
 
 		_, err = db.NamedExec(`INSERT INTO person (first_name, last_name, email)
-			VALUES (:first_name, :last_name, :email) `, slsMap)
+			VALUES (:first_name, :last_name, :email) ;--`, slsMap)
```
jmoiron, line 221: Is this extra comment necessary?

abraithwaite: The comment here is exercising the regex in the sense that anything after the closing paren is ignored.

I could update it to something more clear like `ON CONFLICT` perhaps?  I imagine that most people are using that behavior moreso than random junk at the end.

jmoiron: Happy with the comment, just did not want to add some extra load bearing requirement here.  I think it's clear from the tests later on that these are testing some boundary conditions.

jmoiron:
Tests looked good, tagged as v1.3.2

PR #720, opened 2021-04-08
fix: bulk insert regular expression is too strict

fifsky:
This regularization is too strict, this leads to the failure of all the following cases:

```sql
insert into users(a,b) values(:a,:b)
insert into users(a,b) values (:a,:b)
insert into users(a,b) Values (:a,:b)
insert into users(a,b) VALUES(:a,:b)
```

So, I modified the regular expression, compatible with lowercase and space

### DIFF
--- a/named.go
+++ b/named.go
@@ -224,7 +224,7 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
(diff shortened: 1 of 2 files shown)


jmoiron:
:facepalm: went and made almost this exact change (minus a test) when you commented on it, would have merged if I'd seen the PR available.  For anyone else, this is tagged v1.3.3

Issue #694, opened 2021-01-29
Bulk insert assumed that the placeholders will always be at the end


MichaelBui:
Current regex only accept placeholders at the end of the query:
```
var valueBracketReg = regexp.MustCompile(`\([^(]*.[^(]\)$`)
```

This doesn't work with `INSERT .... ON DUPLICATE KEY UPDATE....`

kazz187:
same
https://github.com/jmoiron/sqlx/issues/657

aarengee:
https://github.com/jmoiron/sqlx/pull/667#issuecomment-787731678

powerman:
This is a regression since 18 Nov 2020 b63be72. Actually this issue is tracked in a lot of places: https://github.com/jmoiron/sqlx/issues/505#issuecomment-780360639, https://github.com/jmoiron/sqlx/issues/612#issuecomment-769936281, #690, #712, #705, https://github.com/jmoiron/sqlx/pull/667#issuecomment-787731678, #657...

jmoiron:
This should work after #718  / v1.3.2, please reopen if you are still having issues.

Al2Klimov:
Wanted to manifest this issue's fix by https://github.com/Al2Klimov/sqlx/commit/a3897d3439c167a9e6f7b3cafb8e8e51d831dd36, but the tests fail:

```
➜  sqlx git:(feature/mysql-upsert) go test -race ./...
Disabling PG tests:
    dial tcp [::1]:5432: connect: connection refused
Disabling MySQL tests:
    invalid DSN: missing the slash separating the database name--- FAIL: TestFixBounds (0.00s)
    --- FAIL: TestFixBounds/named_syntax_w/_upsert (0.00s)
        named_test.go:375: mismatched results
    --- FAIL: TestFixBounds/mysql_syntax_w/_upsert (0.00s)
        named_test.go:375: mismatched results
FAIL
FAIL	github.com/jmoiron/sqlx	0.474s
ok  	github.com/jmoiron/sqlx/reflectx	(cached)
ok  	github.com/jmoiron/sqlx/types	(cached)
FAIL
➜  sqlx git:(feature/mysql-upsert)
```

@jmoiron TL;DR: it's still not fixed, but I seem not to have the permission to re-open this one.

CC @lippserd

abraithwaite:
Interesting.  I noticed you changed the regex slightly, but not in a way which would cause issues for those tests.

[...]

PR #721, opened 2021-04-08
named: support mysql 5.7 and earlier syntax

abraithwaite:
Addresses the concerns raised by https://github.com/jmoiron/sqlx/issues/694#issuecomment-815636257


abraithwaite:
Looks like this breaks some existing behaviors.  I don't have time to investigate right now but might have an opportunity next week.  @al2Klimov if you want to take a stab before then, feel free!

PR #726, opened 2021-04-16
Mod. defer close if err != nil

kubosuke:
# description

in (*NamedStmt)Select(),
I want to exec rows.Close() if error != nil, because if error occurs, connection is not returned to pool immediately.


jmoiron:
Isn't rows `nil` in this case?  I can't see how this doesn't result in a panic.

kubosuke:
> Isn't rows nil in this case? I can't see how this doesn't result in a panic.

yes, it's my misunderstanding, I'm sorry.
I'll close this PR.

Issue #722, opened 2021-04-09
unable to form bulk upsert query


lenonqing:
in v 1.3.3
```
query := "INSERT INTO TABLE_VALUES (`tid`, `uid`, `count`, `deposit`, `ver`) VALUES (:tid, :uid, :count, :deposit, :ver) ON DUPLICATE KEY UPDATE `tid`=VALUES(`tid`), `uid`=VALUES(`uid`), `count`=VALUES(`count`), `deposit`=VALUES(`deposit`), `ver`=VALUES(`ver`)"
args := []map[string]interface{}{
        {"tid": 1, "uid": 1, "count": 1, "deposit": 0, "ver": 0},
        {"tid": 2, "uid": 1, "count": 1, "deposit": 0, "ver": 0},
    }
query, argsOut, err := sqlx.Named(query, args)
_, err = db.ExecContext(ctx, query, argsOut...)
```
got `sql: expected 5 arguments, got 10` error.

gurza:
@lenonqing you can call bulk upset with `[]map[string]interface{}` directly. It works if you need quick solution.

```golang
q := "INSERT INTO role(type,value) VALUES (:type, :value)"

pp := []map[string]interface{}{
	{"type": "type1", "value": "val1"},
	{"type": "type2", "value": "val2"},
	{"type": "type3", "value": "val3"},
}

res, err := db.NamedExec(q, pp)
```

colinrgodsey:
it'd be nice to fix this. if you put anything after the `VALUES(...)` segment, the bulk insert seems to fail. This makes it impossible to use any `ON CONFLICT ...` forms of the insert.

QuangTung97:
> Any update on this?

I think version >= 1.3.4 should has already fixed this

lenonqing:
yes, version >= 1.3.4 is ok.

Odas0R:
I can confirm, this works on v1.3.5!

Maybe close this issue?

SenselessA:
I got the same problem with 1.3.5 version

yzinkovets:
same problem on `1.4.0`

Issue #701, opened 2021-02-18
Batch insert using map seems to still not work


saantiaguilera:
Hi, I'm using the master branch since it has the fix of batch inserts on named arguments. Still, when using a map it seems to still fail:

```go
_, err := tx.NamedExec(
    "INSERT INTO state (state_id, name, created_at) VALUES (:id, :name, :time) ON DUPLICATE KEY UPDATE state_id=state_id",
    []map[string]interface{}{
        {"id": "MX-MEX", "name": "Ciudad de México", "time": time.Now()},
	{"id": "BR-SP", "name": "São Paulo", "time": time.Now()},
    },
)
```

Seems to fail with error: `sql: expected 3 arguments, got 6`

Any idea why?

hyperhtml:
Nothing to add here, but I just ran into this exact issue as well.    

I can get the above query to run if you omit `ON DUPLICATE KEY UPDATE state_id=state_id`.   However, I also need that same clause in my query.

alexshtin:
I have same issue. Pinned to `1.2.0` for now.

jmoiron:
Hesitantly going to call this fixed with #734, which is tagged with `1.3.4`

Issue #695, opened 2021-02-04
Prepared statements broken when using "Named" + "In"


samber:
Hi @jmoiron and thanks for your work on this lib!

On the following example: http://jmoiron.github.io/sqlx/#namedParams, you show how to combine Named params and IN, in a single statement.

This is broken since -at least- sqlx@1.3.0 and on master branch.

Example with the following table:

```sql
CREATE TABLE (foo text, bar text);
```

The following code:

```go
type Test struct {
  Foo string `db:"foo"`
  Bar string `db:"bar"`
}

test := []Test{
  {Foo: "hello", Bar: "world 1"},
  {Foo: "hello", Bar: "world 2"},
}

query := "INSERT INTO test (foo, bar) VALUES (:foo, :bar);"
query, args, err := sqlx.Named(, arg)
query, args, err := sqlx.In(query, args...)
query = db.Rebind(query)
_, err := db.Query(query, args...)

fmt.Println(err.Error())
// "pq: got 4 parameters but the statement requires 2"

fmt.Println(query)
// "INSERT INTO test (foo, bar) VALUES (?, ?);"
fmt.Println(args)
// [hello world 1 hello world 2]
```

Any idea?

kgritesh:
Facing the same issue?

pete-woods:
Another person suffering the same issue here

gnuletik:
v1.3.3 seems to fix the issue :)

jmoiron:
If it wasn't fixed in v1.3.3, there's _another_ fix for these types of issues in v1.3.4

0x457:
batch insert for support oracle? I got the same error

PR #734, opened 2021-04-27
Fix Bulk Insert: Single Column and MySQL earlier upsert syntax

QuangTung97:
This change handle the case of using single column in "VALUES(...)".
And allow using VALUES in "INSERT ... ON DUPLICATE KEY UPDATE ..." kind of queries,  e.g issue https://github.com/jmoiron/sqlx/issues/722
Possible fixes for https://github.com/jmoiron/sqlx/issues/701 https://github.com/jmoiron/sqlx/issues/695 https://github.com/jmoiron/sqlx/issues/694

### DIFF
--- a/named.go
+++ b/named.go
@@ -245,7 +245,7 @@ func findMatchingClosingBracketIndex(s string) int {
 func fixBound(bound string, loop int) string {
 	loc := valuesReg.FindStringIndex(bound)
 	// defensive guard when "VALUES (...)" not found
-	if len(loc) < 1 {
+	if len(loc) < 2 {
 		return bound
 	}
 
--- a/named.go
+++ b/named.go
@@ -224,13 +224,13 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valueBracketReg = regexp.MustCompile(`(?i)VALUES\s*(\([^(]*.[^(]\))`)
+var valueBracketReg = regexp.MustCompile(`(?i)VALUES\s*(\((?:[^(]|\([^(]*\))*\))`)
 
 func fixBound(bound string, loop int) string {
 
 	loc := valueBracketReg.FindAllStringSubmatchIndex(bound, -1)
-	// Either no VALUES () found or more than one found??
-	if len(loc) != 1 {
+	// defensive guard when "VALUES (...)" not found
+	if len(loc) < 1 {
 		return bound
 	}
 	// defensive guard. loc should be len 4 representing the starting and
--- a/named.go
+++ b/named.go
@@ -224,29 +224,47 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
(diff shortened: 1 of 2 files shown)


jakeaglass:
This solves issues I was having with the new batch insert feature, which was generating statements with unbalanced parentheses, eg `INSERT (id) VALUES (coalesce(:a,:b))` compiled to `INSERT (id) VALUES (coalesce(1,2)`

Many thanks @QuangTung97

w1ck3dg0ph3r:
Thanks @QuangTung97!
Was about to create similar PR that does away with this brittle regex)

By the way, this should also fix some additional cases:
- Functions inside values: `INSERT INTO foo (a,b,c,d) VALUES (:name, cast(:age as int), :first, :last)`
- Some strange regex indentation failure I was having on multi-line queries (regex failed on some levels of indentation and not the others, go figure).

Could you add those test cases as well, please? Here's the gist of the ones I added: https://gist.github.com/w1ck3dg0ph3r/cfbce1590fdce1ce54a179b6cb74f6a4.

QuangTung97:
> Thanks @QuangTung97!
> Was about to create similar PR that does away with this brittle regex)
> 
> By the way, this should also fix some additional cases:
> 
> * Functions inside values: `INSERT INTO foo (a,b,c,d) VALUES (:name, cast(:age as int), :first, :last)`
> * Some strange regex indentation failure I was having on multi-line queries (regex failed on some levels of indentation and not the others, go figure).
> 
> Could you add those test cases as well, please? Here's the gist of the ones I added: https://gist.github.com/w1ck3dg0ph3r/cfbce1590fdce1ce54a179b6cb74f6a4.

Thanks. I added test for the multiple lines query case. But the case of function calls might be redundant with the test case `two level depth function call` I already added.

jmoiron:
This looks great;  this regex has seen a lot of authors and is difficult to get right, but this looks great and more tests are also welcome.  Going to cut a new dot release w/ this sometime this weekend, triaging more in the backlog.

named.go
```diff
@@ -224,29 +224,47 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valueBracketReg = regexp.MustCompile(`(?i)VALUES\s*(\([^(]*.[^(]\))`)
+var valuesReg = regexp.MustCompile(`\)\s*(?i)VALUES\s*\(`)
```
w1ck3dg0ph3r, line 227: Should we also care for inserts without an explicit column list here, e.g. `insert into tblname values (...)`?

QuangTung97: Oh. Thanks for pointing out. But I dont find a good way to both satisfying that case and the case when the name of table containing suffix: "values".

w1ck3dg0ph3r: I think `\W+(?i)VALUES\s*\(` should handle most cases, including:
```
INSERT INTO foo_values (a,b,c,d) VALUES (?, ?, ?, ?)
INSERT INTO foo_values VALUES (?, ?, ?, ?)
INSERT INTO `values`VALUES (?, ?, ?, ?)
```
While generally being a bad idea, I have no doubt that somebody does implement it, given such a wide adoption of sqlx)

PR #758, opened 2021-07-21
Add oracle to defaultBinds

abeMedia:
Add `oracle` to defaultBinds for use with https://github.com/sijms/go-ora (the only oracle driver that's pure go).


BenKnigge:
this was added to my fork here https://github.com/BenKnigge/sqly

PR #760, opened 2021-07-30
added support for named WITH query

imle:
Added support for WITH queries which still use values, but have a different prefix than is checked for by the current regex.


ccoVeille:
fixed by @mrj0 in his fork

- https://github.com/Vinovest/sqlx/pull/7

Issue #744, opened 2021-06-03
Make Select reset slice length


cespare:
sqlx's `Select` functions take a destination slice and fill it in. It's not documented, but they simply append to the slice; if it already had some elements, they are left alone.

Here's a demo:

```
package main

import (
	"encoding/json"
	"fmt"
	"log"

	"github.com/jmoiron/sqlx"
	_ "github.com/lib/pq"
)

func main() {
	dsn := "host=localhost dbname=postgres"
	db := sqlx.MustConnect("postgres", dsn)

	db.MustExec(`
DROP TABLE IF EXISTS sqlx_test;
CREATE TABLE sqlx_test ( x text );
INSERT INTO sqlx_test (x) VALUES ('a'), ('b');
`)

	var s []string
	if err := db.Select(&s, "SELECT x FROM sqlx_test"); err != nil {
		log.Fatal(err)
	}
	if err := db.Select(&s, "SELECT x FROM sqlx_test"); err != nil {
		log.Fatal(err)
	}
	fmt.Println(s) // prints [a b a b]

	if err := json.Unmarshal([]byte(`["json"]`), &s); err != nil {
		log.Fatal(err)
	}
	fmt.Println(s) // prints [json]
}
```

The behavior is a little surprising, and it's also the opposite of how [JSON unmarshaling behaves](https://golang.org/pkg/encoding/json/#Unmarshal):

> To unmarshal a JSON array into a slice, Unmarshal resets the slice length to zero and then appends each element to the slice.

Below I'll describe why this causes problems for us.

I propose that we change the behavior of `Select` to first reset the slice length, just like `json.Unmarshal`.

It seems possible that people depend on this behavior due to Hyrum's Law, so if you want to avoid breaking them, I guess a slightly different change would be needed. Perhaps a DB-level bool option?

If there's a change you're willing to accept here, we can send a PR.

---

Here's where this causes problems for us.

We have some applications that make heavy use of PostgreSQL using the serializable isolation level. In this mode, all queries (including read queries) need to be retried on serialization failures. The way this looks is that we have some helper functions which take a querying callback and run it inside a retry loop. Here's a slightly simplified example:

```
func QueryWithTxRetries(db *sqlx.DB, fn func(*sqlx.Tx) error) error {
	tryTxFn := func() error {
		tx, err := db.Beginx()
		if err != nil {
			return err
		}
		defer tx.Rollback()
		if err := fn(tx); err != nil {
			return err
		}
		return tx.Commit()
[...]

PR #767, opened 2021-08-19
Make Select reset slice length

gayoungpark:
Fixes issue #744.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -878,9 +878,10 @@ func structOnlyError(t reflect.Type) error {
 }
 
 // scanAll scans all rows into a destination, which must be a slice of any
-// type.  If the destination slice type is a Struct, then StructScan will be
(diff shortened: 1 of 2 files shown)


cespare:
@jmoiron WDYT about merging this? We've been using it in production for a while and it has been a very useful change for us.

jmoiron:
v1.3.5

cespare:
Thanks @jmoiron !

cpuschma:
Thank you @jmoiron for still caring about the repository ❤️

benCoomes:
👋 Hi @jmoiron! I recently updated my application to 1.3.5 and noticed my tests were failing due to this change. I guess I'm one of the victims of Hyrum's law. 🥲 

Since this is a breaking change, would you consider making the behavior opt-in? Or, at the very least, updating the major version. Folks won't be expecting a change like this on a patch update.

PR #770, opened 2021-09-02
add a column name mapper func

0xor1:
it may be generally helpful to process column names when mapping, similar to how field names and tag names can be preprocessed in the mapper. This is an optional additive change.

having to add a db or json tag to every struct field just to lower case the first character of the field feels very redundant, when we could just ToLower both the field name and the column name to match.


dlsniper:

Issue #703, opened 2021-02-21
"Unsafe" as a backwards compat method


aep:
Hi,

i'm confused why Unsafe is called unsafe. The naming suggests i'm misunderstanding fundamentally how you're supposed to retain backwards compatibility when adding fields to a database.

Without Unsafe, when adding a field, older database clients that dont know the field yet will fail. So you must first upgrade all  clients and THEN migrate, which sounds backwards?

the alternative "correct" solution would be to explicitly name every field in the struct AND the query, but then what's the added benefit over just using stdlib sql?

vm-001:
Lots of applications run `migrations up` first and then restart the application. SQLs `select * from X` would fail after it saw any unknown column. I'm also looking for a global configuration to set unsafe always to true without manually calling.Unsafe()

PR #789, opened 2021-12-07
UnsafeLogged added to report missing fields

mrabe89:
Hi, - We use your great sqlx library in our projects.
I added an optional logging functionality to your unsafe skips to report if a destination is missing, but to still continue.
I'm not sure if this is the best way to do that, but if not please provide me with feedback and I will work your suggestions in.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 
+	"io"
 	"io/ioutil"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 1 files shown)


gregwebs:
wow, I was just thinking about making this change and now I see others have needed it for years.
The existing error is often helpful, but not for a zero-downtime deploy- in that case this error needs to be turned off to avoid further complicating the deployment. This issue has been [raised already](https://github.com/jmoiron/sqlx/issues/703).

Given the simple logging needs here, I think it makes sense to use the writer instead of an actual Log instance. Using slog is tempting, but not sure if there is a way to make that backwards compatible with older versions of Go.

Issue #344, opened 2017-07-27
Generic *sqlx.DB or *sqlx.Tx?


mewben:
Hi, is there a way I could accept a parameter which is either `sqlx.DB` or `sqlx.Tx`?

I have this function wherein I have to pass `*sqlx.DB`: 
```go
func GetUser(db *sqlx.DB, username string) ...
```

There's no problem with this, until I create a new Transaction and inside that transaction, I need to call `GetUser`. So I ended up implementing 2 functions to provide the same result...
```go
func GetUser2(tx *sqlx.Tx, username string) ...
```

Is there a way I can make this into 1 function? I tried this:
```go
func GetUser(db *sqlx.DB, tx *sqlx.Tx, username string) (*User, error) {
  if tx != nil {
    // use tx
  } else {
    // use db
  }
  ...
}
```
but is this okay? I wish there is something like this:
```go
func GetUser(db *sqlx.DBorTxSomething, ...
```
:)

jhngrant:
@mewben This is consistent with *sql.Tx and *sql.DB in the standard library.

Over in golang-nuts, @davecheney suggests defining your own interface.

See: https://groups.google.com/forum/#!topic/golang-nuts/yLg6IWbwPPw

alhails:
I've achieved this by creating an interface that covers the sqlx methods I need and then creating two implementations that wrap sqlx.DB and sqlx.Tx respectively.  Then your GetUser method can accept that interface instead of the concrete types.

eexit:
Hey, you need to create a common interface that is used by both `*sqlx.DB` and `*sqlx.Tx`.
Example:

```go
type DB interface {
	Querier
	Begin() (Tx, error)
	Close() error
}

type Tx interface {
	Querier
	Commit() error
	Rollback() error
}

type Querier interface {
	QueryRow(query string, args ...interface{}) Row
	Select(dest interface{}, query string, args ...interface{}) error
	NamedQuery(query string, arg interface{}) (Rows, error)
	NamedExec(query string, arg interface{}) (sql.Result, error)
}


func GetUser(q *Querier, username string) {}
```

mewben:
Ahhh.. I get it now.. Thank you all...

tj:
Maybe it's just me but I'd still love to see this, bit of a pain to need yet-another layer

wubin1989:
> Hey, you need to create a common interface that is used by both `*sqlx.DB` and `*sqlx.Tx`.
> Example:
> 
[...]

PR #809, opened 2022-05-09
Add Queryable interface

BlakeWilliams:
I came across https://github.com/jmoiron/sqlx/issues/344 and thought it may be
useful to open a PR proposing this change since it looks like a few folks could
find it useful.

---

When writing applications it's useful to pass a sqlx.DB and a sqlx.Tx
interchangeably so that you can compose functions that are runnable in
isolation or as part of a transaction.

This introduces the Queryable interface, which includes the common
exportable methods shared between sqlx.DB and sqlx.Tx so users of the
package don't have to implement it themselves.

This also adds tests that validate any new shared methods are
added to the interface.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -1,6 +1,7 @@
 package sqlx
 
 import (
+	"context"
 	"database/sql"
 	"database/sql/driver"
 	"errors"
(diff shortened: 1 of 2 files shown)


FikriRNurhidayat:
This is actually very helpful, even tho  it hasn't being merged yet, but it does help me to implement Unit of Work

1mi0:
I think it would make more sense to split the giant interface into a bunch of smaller ones. I do completely agree that having the common methods to `sqlx.Tx` and `sqlx.DB` is really useful. But, I also think that a giant interface is hard to mock and deal with.

The Queryable interface carries a lot more functionality than is usually needed(in the use cases that I've seen). Instead, having interfaces like: `Getter`, `NamedExecer`, `Selector`, and their Context alternatives, makes more sense for small CRUD operations, where you also need to mock test them.

I think that composing Queryable out of smaller interfaces would also tie in better with the current style. There already are some interfaces like that.
https://github.com/jmoiron/sqlx/pull/809/files#diff-e6a1c006245175d26ca4650b0bf4fc8484a43de15056a1634e83e825193e9018R244-R248

dlsniper:
I haven't had a chance to discuss this with @ardan-bkennedy; that's why the label is set to `needs testing`.

I think we don't need this interface in the package at all. This should be an application concern to define/use. That way, the application establishes the functionality it expects. We should ensure that both `sqlx.Tx` and `sqlx.DB` always have identical signatures for the shared methods to make it possible to define such interfaces. In this case, I suggest creating this interface but not exporting it.

ardan-bkennedy:
This interface already exists and I use it today to substitute a DB for a Tx when needed. 

sqlx.ExtContext

I haven't looked too closely to this issue, but why won't this interface work for you?

mrj0:
> This interface already exists and I use it today to substitute a DB for a Tx when needed.
> 
> sqlx.ExtContext
> 
> I haven't looked too closely to this issue, but why won't this interface work for you?

This type is still missing several methods, like Get, GetContext and SelectContext. Adding Queryable would be nicer than library users needing to dig for and maintain definitions. :+1: for pr

```
type SqlBase interface {
	sqlx.Ext
	sqlx.ExtContext
}
```

Issue #796, opened 2022-01-24
Batch update support


TmLev:
Does sqlx support batch update query? If so, how one should use it correctly?

Swiftwork:
The syntax should simply be:

```sql
UPDATE rows AS r
SET r.column1 = v.column1,
    r.column2 = v.column2
FROM (VALUES
    (:id, :column1, :column2)
) AS v(id, column1, column2)
WHERE r.id = v.id
```

However, this doesn't currently work in `v1.3.4` due to the current 
regex requiring a closing bracket before values
https://github.com/jmoiron/sqlx/blob/92bfa368c21aafd33626444a47b2f99be2432623/named.go#L227

This was discussed by @QuangTung97 and @w1ck3dg0ph3r when it was initially implemented https://github.com/jmoiron/sqlx/pull/734/files#r632982543 and possible changes required to mitigate this.

Also worth noting is that update doesn't automatically coerce type https://www.postgresql.org/docs/14/sql-values.html and you might have to be explicitly cast the types. More info on this can be found in the comments on answer https://stackoverflow.com/a/18799497. 

> An example with a date: ... from (values ('2014-07-21'::timestamp, 1), ('2014-07-20', 2), ...

A final caveat is the casting using `::` probably won't work with named queries https://github.com/jmoiron/sqlx/issues/91, but a workaround could be `CAST('2014-07-21' AS timestamp)`.

chrichts:
For anyone else who gets here, another temp workaround until this is fixed is doing something like:

```
INSERT INTO rows (column1 ,column2 ) VALUES (:column1, :column2) ON DUPLICATE KEY UPDATE column1=:column1
		
```

dan-r95:
According to [this comment](https://github.com/jmoiron/sqlx/issues/722#issuecomment-834673981)   on a named batch update every variable after the VALUES part produces errors.

PR #811, opened 2022-05-13
fix: bindArray on SELECT from VALUES list

Phoenix1708:
- `fixBound` currently assumes named var in VALUES is only from INSERT
- `bindArray `is not populating the right number of placeholders

Ideally the regex should be something like: `(?<!(?i)into)\W+(?i)VALUES\s*\(` but Golang regex doesn't support look around

### DIFF
--- a/named.go
+++ b/named.go
@@ -224,7 +224,7 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valuesReg = regexp.MustCompile(`\)\s*(?i)VALUES\s*\(`)
+var valuesReg = regexp.MustCompile(`[\)|\(]\s*(?i)VALUES\s*\(`)
 
 func findMatchingClosingBracketIndex(s string) int {
 	count := 0
--- a/named_test.go
+++ b/named_test.go
@@ -392,6 +392,18 @@ func TestFixBounds(t *testing.T) {
 			loop:   2,
 		},
 		{
+			name:   `table named "values"`,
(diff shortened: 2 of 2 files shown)


haton14:
WDYT about merging this? @jmoiron 
Together, this issue seems to be solved.https://github.com/jmoiron/sqlx/issues/796

masv87:
@jmoiron We are suffering a lot from this bug, please merge.

amirhood:
@jmoiron +1
Our destiny depends on this fix. Please review and merge if possible. Thanks.

tylerstillwater:
I'd also love to see this merged so #796 could be solved.

named.go
```diff
@@ -224,7 +224,7 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valuesReg = regexp.MustCompile(`\)\s*(?i)VALUES\s*\(`)
+var valuesReg = regexp.MustCompile(`[\)|\(]\s*(?i)VALUES\s*\(`)
```
jvnuw, line 227: I see test cases justifying `)` and `(` but not `|`. Unless anyone wants to add a test demonstrating the usefulness of allowing a leading `|`, maybe this PR could get unblocked by removing that option?

```suggestion
var valuesReg = regexp.MustCompile(`[)(]\s*(?i)VALUES\s*\(`)
```

Note there is no need to escape parentheses within a character class, since capture groups cannot exist inside character classes.

ikholmatov:
any news ?

brentd:
@jmoiron this would be a huge improvement to the library for myself and many people. Please consider merging this PR and thank you for the library!

Issue #162, opened 2015-08-25
StructScan rows after outer join


Perlence:
Is there a way to simply embed e.g. "right" table into "left" table and make "LEFT OUTER JOIN"?
Columns on the right can be nil, but I don't feel that copy-pasting type definition replacing types with their Null counterparts is good idea.

```go
type EmployeeDepartment struct {
  Employee
  Department
}

type Department struct {
  DepartmentID uint64
  DepartmentName string
}

type Employee struct {
  LastName string
  DepartmentID int64
}

func join(db *sqlx.DB) error {
  var eds []EmployeeDepartment
  err := db.Select(&eds, `SELECT *
    FROM employee 
    LEFT OUTER JOIN department
    ON employee.DepartmentID = department.DepartmentID;`)
  if err != nil {
    // err is not nil: "Scan error on column index 2: unsupported driver Scan pair: <nil> -> *string".
    return err
  }
  return nil
}
```

I tried to perform two separate scans and put them together:

```go
func join(db *sqlx.DB) error {
  var eds []EmployeeDepartment
  rows, err := db.Queryx(`SELECT *
    FROM employee 
    LEFT OUTER JOIN department
    ON employee.DepartmentID = department.DepartmentID;`)
  for rows.Next() {
    var e Employee
    err = rows.StructScan(&e)
    if err != nil {
      return err
    }
    var d Department
    // Here StructScan panics: "reflect: Field index out of range"
    err = rows.StructScan(&d)
    if err != nil {
      // Handle "unsupported driver Scan pair" and move along.
    }
    eds = append(eds, EmployeeDepartment{Employee: e, Department: d})
  }
  return nil
}
```

nubbel:
+1

I would love to have sqlx allow this syntax:
```go
type EmployeeDepartment struct {
  Employee
  *Department
}
```

schickling:
+1

Current workaround involves a null-object in the database which feels **really wrong**

schickling:
@jmoiron are there any plans to support this in the near future?

jriquelme:
I'm facing the same situation. Would be nice to allow @nubbel proposal.

kliron:
+1 Same here.

v4n:
A possible solution woudl be to add multiple tag values support to SQLX (`database:"id,tablenname_id"`) so we could do this with the ALIAS sql syntax:
```
type Department struct {
  ID int `database:"id,department_id"`
}

type Employee struct {
[...]

PR #847, opened 2023-01-23

ntbosscher:
Nested structs are now only instantiated when one of the database columns in that nested struct is not nil. This allows objects scanned in left/outer joins to keep their natural types (instead of setting everything to NullableX).

Example:
```sql
select house.id, owner.*,
from house
left join owner on owner.id = house.owner
```

### Before

``` golang
type House struct {
 ID      int
 Owner   *Person // if left join gives nulls, Owner is set
}

type Owner struct {
 ID sql.NullInt64 // make nullable columns even tho if table doesn't have those columns nullable
}
```

### After

``` golang
type House struct {
 ID      int
 Owner   *Person // if left join gives nulls, Owner will be nil
}

type Owner struct {
 ID int    // no need to set this to sql.NullInt
}
```

### DIFF
--- a/reflectx/reflect.go
+++ b/reflectx/reflect.go
@@ -7,8 +7,11 @@
 package reflectx
 
 import (
+	"database/sql"
+	"fmt"
 	"reflect"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 )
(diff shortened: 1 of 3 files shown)


ntbosscher:
A possible solution for #162

ntbosscher:
One trade off is that for nested structs, we end up doing most of the work in [sql.convertAssignRows](https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/database/sql/convert.go;drc=ba913774543d7388b7bb1843fc7c1b935aebedda;bpv=0;bpt=1;l=490) twice

This also further complicates reflectx.

However, I think the functionality is more natural. With our modeling we typically do things like the following.

```golang
type User struct {
    ID int
    Name string
}

type Plan struct {
    ID int
    CreditsPerMonth int64
}

type UserPlan struct {
   User User
   Plan *Plan
}
```

This maps directly to database tables. Doing things like setting Plan's properties to nullable fields feels weird and out of place.

ntbosscher:
Happy to make any adjustments if you see a better way to do this.

Omar-V2:
This makes a lot of sense, I ran into issues with the existing behaviour recently. I don't like that I have to make a bunch of fields Nullable on existing structs just so they can be properly scanned when performing a left join.  Any updates from the team if this can get merged? @jmoiron? Thanks!

maranqz:
@jmoiron, Could you check and merge it? Thanks. 🙏

geeeeeeeeek (approved):
Tested locally and it works with multi outer joins. Vote to merge this PR.

geeeeeeeeek (changes requested):
Wait it doesn't work with pointer types: `don't know how to parse type int64 -> **uint64`

```
type AccountRow struct {
	Account     `db:"account"`
	OptionOrder `db:"option_order"`
}
type OptionOrder struct {
	ID        uuid.UUID `db:"id"`
	AccountID *uint64   `db:"account_id"`
}
```

PR #852, opened 2023-03-19
Add example for arbitrary columns

Rmarken5:
During the development of my project, I came across the need to execute a multi-table join using sqlx.This resulted in arbitrary column names in the result set. I aimed to scan the resulting rows into a slice of structs containing embedded structs with `db` tagging.

Unfortunately, I was unable to find any documentation for such a scenario. (I did see the "known issues" section of the readme)
After forking sqlx , replacing the package in my go mod file with my local version, debugging, and experimenting, I was able to achieve the desired result after a considerable amount of time.

In the interest of helping others who may encounter similar issues, I suggest extending the readme usage with an example directory that provides additional examples of not-so-straightforward scenarios. This will be beneficial for those adopting this package into their projects.


dlsniper:
We are sorting the opened issues and pull requests. We'll look at the example provided and merge it if everything works as expected..

Issue #857, opened 2023-04-12
Scanning multiple result sets using Rows (NextResultSet keeps outdated cache).


kamilkoduo:
When reading multiple result sets from sqlx.Rows using StructScan, some information obtained using costly reflect operations is cached for performance optimisation. However, different result sets in a single sqlx.Rows object might have different structures, so cache should be invalidated and rebuild after each NextResultSet call. Now this method is not overwritten, so old cache interferes with new structures blocking struct scan after first result set.

PR #747, opened 2021-06-21
added NextResultSet implementation

kmpm:
Trying to revive the solution for issue #380 

Shamelessly used the stuff that @basvanbeek created in pr #453 but added some testing with postgres.
This would also work with mssql but I didn't want to mess with the test environment.

PR #858, opened 2023-04-12
Invalidate cache on Rows.NextResultSet call

kamilkoduo:
When reading multiple result sets from sqlx.Rows using StructScan, some information obtained usCCCing costly reflect operations is cached for performance optimisation.
However, different result sets in a single sqlx.Rows object might have different structures, so cache should be invalidated and rebuild after each NextResultSet call. Now this method is not overwritten, so old cache interferes with new structures after first result set scan.

Relates #857


cornelk:
Thanks for the fix! There is another PR open for this issue: https://github.com/jmoiron/sqlx/pull/747 which includes a unit test.

@jmoiron any chance to get this or the other fix merged?

batazor:
@jmoiron Yeah, that's a real problem, it would be cool if you could merge one of these solutions

dlsniper:
We are currently evaluating all the opened PRs.
We'll take a look at this shortly and see if this, #747 or another, will be a good solution for this problem.

louis77:
I just found this issue and wonder if anyone would merge it? I tested the proposed changes in the referenced commit, and it works for all my queries.

PR #900, opened 2023-11-01

geeeeeeeeek:
Based on https://github.com/jmoiron/sqlx/pull/847, but fixed an issue that optional pointer fields are not handled correctly.

```
Nested structs are now only instantiated when one of the database columns in that nested struct is not nil. This allows objects scanned in left/outer joins to keep their natural types (instead of setting everything to NullableX).

Example:

select house.id, owner.*,
from house
left join owner on owner.id = house.owner
Before
type House struct {
 ID      int
 Owner   *Person // if left join gives nulls, Owner is set
}

type Owner struct {
 ID sql.NullInt64 // make nullable columns even tho if table doesn't have those columns nullable
}
After
type House struct {
 ID      int
 Owner   *Person // if left join gives nulls, Owner will be nil
}

type Owner struct {
 ID int    // no need to set this to sql.NullInt
}
```

### DIFF
--- a/reflectx/reflect.go
+++ b/reflectx/reflect.go
@@ -7,8 +7,11 @@
 package reflectx
 
 import (
+	"database/sql"
+	"fmt"
 	"reflect"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 )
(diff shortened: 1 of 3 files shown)


dlsniper:

taleeus:
This would be insanely useful 🙏🏻

AlexandrePhilibert:
Would love for this to be merged. I have been using sqlx with these changes for some time as it is a feature I find really useful.

kszafran:
@dlsniper This issue is precisely what forced me to ditch sqlx many years ago. Now I ran into it again in another project... Hoping to see this merged soon 🤞🏻

kszafran:
I tried this change in my project, and there are definitely breaking changes. The code copies from stdlib `convertAssign`, but it's not fully compatible with it. For example, without this change scanning an `int64` value into `string` field works fine, but with this change it fails. I'm looking at the code trying to figure out I could address this issue.

kszafran:
@dlsniper @ardan-bkennedy I created another followup PR: https://github.com/jmoiron/sqlx/pull/950. It addresses incompatibility issues, adds tests, and is much shorter. It would be great if you could take a look at it!

PR #909, opened 2024-02-01
Upgrade dependencies

dlsniper:
This PR upgrades the dependencies to their latest versions.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,6 +1,8 @@
-github.com/go-sql-driver/mysql v1.6.0 h1:BCTh4TKNUYmOmMUcQ3IipzF5prigylS7XXjEkfCHuOE=
-github.com/go-sql-driver/mysql v1.6.0/go.mod h1:DCzpHaOWr8IXmIStZouvnhqoel9Qv2LBy8hT2VhHyBg=
-github.com/lib/pq v1.2.0 h1:LXpIM/LZ5xGFhOpXAQUIMM1HdyqzVYM13zNdjCEEcA0=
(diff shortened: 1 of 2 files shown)


go.sum
```diff
@@ -1,6 +1,12 @@
 github.com/go-sql-driver/mysql v1.6.0 h1:BCTh4TKNUYmOmMUcQ3IipzF5prigylS7XXjEkfCHuOE=
 github.com/go-sql-driver/mysql v1.6.0/go.mod h1:DCzpHaOWr8IXmIStZouvnhqoel9Qv2LBy8hT2VhHyBg=
+github.com/go-sql-driver/mysql v1.7.1 h1:lUIinVbN1DY0xBg0eMOzmmtGoHwWBbvnWubQUrtU8EI=
+github.com/go-sql-driver/mysql v1.7.1/go.mod h1:OXbVy3sEdcQ2Doequ6Z5BW6fXNQTmx+9S1MCJN5yJMI=
```
JRaspass, line 4: This doesn't look like it was `go mod tidy`-ied. I would expect the old versions to go away.

uvulpos:
Is this pr still gonna see the daylight? Because the sqlite3 dependency is already outdated again 👀 
Dependabot or RenovateBot might help

dlsniper:
This should be merged after https://github.com/jmoiron/sqlx/pull/922.
@ardan-bkennedy

dlsniper:
> Is this pr still gonna see the daylight? Because the sqlite3 dependency is already outdated again. It's pending for over a month now again 👀 Dependabot or RenovateBot might help

Hi. I don't think we need the noise from bots for now. We'll try to upgrade the dependencies before package releases, but this is something @ardan-bkennedy and I haven't talked about yet.

PR #914, opened 2024-02-22
Add missing Tx.NamedQueryContext

michaelshobbs:
In an effort to use a postgres `INSERT ... RETURNING` stmt in a transaction, I found that there was a `(*sqlx.Tx).NamedQuery` but not a `(*sqlx.Tx).NamedQueryContext`. This PR attempts to remedy that.


ringmaster217:
@dlsniper I find myself needing this exact same addition, and you seem to be the most active contributor on this project at the moment. What would it take to get some eyes on this and get it merged?

PR #922, opened 2024-04-13
Introduce CircleCI

dlsniper:
With this PR, we introduce the CircleCI as a CI pipeline.

The old configuration, TravisCI, is removed.

A Makefile is also created to allow easier interaction with the different commands available.


dlsniper:
The failures are due to some of the tooling not supporting older versions of Go.

Some other failures are because vulncheck finds issues with the Go version being too old and requiring an upgrade to a newer Go version.

Should we just support the Go 1.20+ in CI?

PR #927, opened 2024-05-28
Add azuresql driver bind

ninjadq:
When accessing azure managed sql server via Azure AD for authentication we need to use [`azuresql` driver](https://github.com/microsoft/go-mssqldb/blob/main/azuread/driver.go#L15)

So the behavior of `azuresql` is the same of `sqlserver`. 
But in our default bind doesn't include that driver name

This PR added it to default bind with AT

### DIFF
--- a/bind.go
+++ b/bind.go
@@ -22,10 +22,10 @@ const (
 )
 
 var defaultBinds = map[int][]string{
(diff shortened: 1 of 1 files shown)


ninjadq:
Hi, 
When I use `azuresql` as driver. The sql will executed with error like 
`mssql: Incorrect syntax near '?'`
This PR can fix it.

yuphing-ong:
This would also fix my issue #928, thanks

dlsniper (approved):

PR #948, opened 2024-10-04
Add support for clickhouse settings in NamedExec's fixBounds

stephennancekivell:
Hi folks, im having trouble with bulk inserts into clickhouse. NamedExec queries are only getting the first value in the array.

This PR fixes the issue by extending the regex to support settings.

Eg this test case fails, with only joe getting inserted, not sally.
```go
func TestInsertNamedBulk(t *testing.T) {
	_, err := db.Exec(schema)
	assert.NoError(t, err)
	defer cleanup(t)

	// Insert data
	_, err = db.NamedExec(
		`insert into users (id, name, age)
		SETTINGS async_insert=1, wait_for_async_insert=1
		values
		(:id, :name, :age)`,
		[]map[string]any{
			{
				"id":   1,
				"name": "joe",
				"age":  30,
			},
			{
				"id":   2,
				"name": "sally",
				"age":  31,
			},
		})
	assert.NoError(t, err)

	users, err := findUsers()
	assert.NoError(t, err)
	assert.Equal(t, []User{
		{ID: 1, Name: "joe", Age: 30},
		{ID: 2, Name: "sally", Age: 31},
	}, users)
}
```

### DIFF
--- a/named.go
+++ b/named.go
@@ -224,7 +224,7 @@ func bindStruct(bindType int, query string, arg interface{}, m *reflectx.Mapper)
 	return bound, arglist, nil
 }
 
-var valuesReg = regexp.MustCompile(`\)\s*(?i)VALUES\s*\(`)
+var valuesReg = regexp.MustCompile(`\)\s*(?i)(?:SETTINGS\s+.*)?\s*VALUES\s*\(`)
 
 func findMatchingClosingBracketIndex(s string) int {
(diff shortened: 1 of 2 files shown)


mrj0:
I think this one would also solve your problem without adding db-specific patterns to the regex: https://github.com/jmoiron/sqlx/pull/962/files

PR #958, opened 2024-12-19
SEC-2706

kasluthra-sec:


mrj0:
This breaks the interface and all users of this library. It also doesn't seem to prevent the most likely source of vulnerabilities, somebody accidentally string cat'ing unsafe data but then wrapping it in safesql because it's required. ```safesql.New(`select ...`+userData)```. The safesql module doesn't do any parsing, just a new type to enforce at compile time. Unless I'm missing something it feels like this should be a fork?

kasluthra-sec:
@mrj0 So Sorry, we were experimenting and I think I had accidentally opened this PR against the wrong upstream. Thanks for your comments, appreciate the insight :)

PR #968, opened 2025-06-13
range over runes in compileNamedQuery func

aachex:


named.go
```diff
@@ -131,7 +131,7 @@ type namedPreparer interface {
 
 func prepareNamed(p namedPreparer, query string) (*NamedStmt, error) {
 	bindType := BindType(p.DriverName())
-	q, args, err := compileNamedQuery([]byte(query), bindType)
```
aachex, line 134: x

mrj0:
Thank you for looking into this! I think if you look at https://github.com/Vinovest/sqlx your problem will be solved there already

PR #978, opened 2025-10-28
feat: Expose scanning any/all without enforcing struct

zacharysierakowski:
## High level

Exposes two functions - `ScanAll` and `ScanSingleRow` that can be used with a `Rows` instance to safely/easily bind to a `dest` that's a struct _or_ a scannable. 

## Use case

The reason for this addition is, I'm writing an wrapper around `*sqlx.DB` that essentially prepends a GTID check into a multi-statement. I'm trying to support `GetContext` and `SelectContext` so that they can be called without any difference to the function signature when using the wrapper (and no need for awareness of wrapper logic at all).

The code I've written so far that's running into this blocker is:
- prepends a separate query to the incoming query
- using `QueryxContext` to get the results of the multi-statement queries
- reading the first result, doing things, then moving onto the second (`NextResultSet`)
- once I'm onto the second, I basically want to bind to `dest` just like the standard `GetContext` would
- [blocker] however, if I use something like `rows.StructScan(dest)`, it panics if the provided `dest` isn't a struct

psuedo-code
```
func (g *DB) GetContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error {
    ...
    multiStatementQuery, combinedArgs := prependQuery(query)
    rows, err := g.d.QueryxContext(ctx, multiStatementQuery, combinedArgs...)
    ...
    defer rows.Close()
    ... 
    rows.Scan(xxx) // read the first query results
    ... 
    rows.NextResultSet() // move to next result set
    ... 
    rows.StructScan(dest) // this is what fails, when a caller passes in a scannable to this wrapper
}
```

With this PR, the end state would be that I can swap my last line (`rows.StructScan(dest)`) to `sqlx.ScanSingleRow(rows, dest, false)`. 

I have the same issue with my `SelectContext` function where i'll need `sqlx.ScanAll`. 

## Details

This PR exposes `scanAll` as a public function. The function has been unmodified, other than updating the casing (so that it's public) and removing the sentence in the comment about the desirability to make it public 🙂 all calls to `scanAll` have been updated to `ScanAll`.

Additionally, this PR exposes a function called `ScanSingleRow` which is a subset of the existing `scanAny` function. It was slightly modified to support `Rows` or `Row`, but should not introduce a functional difference from the existing implementation. 

Neither of these changes are breaking changes. They essentially expose the "scan x" functionality so that they can be called without needing to reflect on the dest before choosing to call something like `StructScan`. 

## Additional Context

If it means anything at all, this change is for code I'm writing @GitHub. I'm hoping not to have to keep a `sqlx` fork just for this delta. If there's feedback in terms of naming, or, if you'd prefer this to somehow be put under a subdirectory/package, I'm happy to iterate here! If there's a way to accomplish this with the existing code, without reinventing a chunk of sqlx reflection/scanning, I'm happy to close this PR.

### DIFF
--- a/sqlx.go
+++ b/sqlx.go
@@ -680,7 +680,7 @@ func Select(q Queryer, dest interface{}, query string, args ...interface{}) erro
 	}
 	// if something happens here, we want to make sure the rows are Closed
 	defer rows.Close()
-	return scanAll(rows, dest, false)
+	return ScanAll(rows, dest, false)
 }
 
 // Get does a QueryRow using the provided Queryer, and scans the resulting row
@@ -746,7 +746,15 @@ func (r *Row) scanAny(dest interface{}, structOnly bool) error {
 		return r.err
 	}
 	defer r.rows.Close()
+	return ScanSingleRow(r, dest, structOnly)
+}
 
+// ScanSingleRow scans a single Row into the dest.
+//
+// It assumes that r is positioned on a valid row.
+// If structOnly is true, an error will be returned if dest is a scannable type (for backwards compatibility with StructScan).
+// Callers are responsible for closing r.
+func ScanSingleRow(r ColScanner, dest interface{}, structOnly bool) error {
 	v := reflect.ValueOf(dest)
 	if v.Kind() != reflect.Ptr {
 		return errors.New("must pass a pointer, not a value, to StructScan destination")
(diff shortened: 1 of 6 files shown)


zacharysierakowski:
@jmoiron do you have some time to review this anytime in the next week or so? 🙏

zacharysierakowski:
It doesn't look like I have permissions to request reviewers, so just going to @ mention a few folks from latest PR merges to see if anyone has the permissions to review/approve for merge.

@dlsniper @ardan-bkennedy @pkieltyka

cli1150 (commented):
🚢

zacharysierakowski:
👋 @jmoiron who's the best point of contact for reviewing this?

johnpreed:
👍 Can't wait to get this merged!

sqlx.go
```diff
@@ -775,11 +783,18 @@ func (r *Row) scanAny(dest interface{}, structOnly bool) error {
...
 	}
 
-	m := r.Mapper
-
+	var m *reflectx.Mapper
+	switch rows := r.(type) {
+	case *Rows:
+		m = rows.Mapper
+	case *Row:
+		m = rows.Mapper
```
jidicula:
```suggestion
	case *Rows, *Row:
		m = rows.Mapper
```

jidicula (commented):
Small nit for switch-case syntax

sqlx.go
zacharysierakowski: Thanks @jidicula!

PR #984, opened 2026-05-12
In: skip ? inside SQL comments and string literals

### LINKED ISSUE #961: Inline comment with a question mark ? results in a  "number of bindVars exceeds arguments" error within sqlx.In
Consider this panicing `main.go` file

```go
package main

import (
	"fmt"

	"github.com/jmoiron/sqlx"
)

func main() {
	query, args, err := sqlx.In(`
		SELECT
			*	-- Should we use field names ?
		FROM students
		WHERE id IN (?)
	`,
		[]int{1, 2, 3},
	)

	if err != nil {
		panic(fmt.Sprintf("Failed to generate query: %v", err.Error()))
	}

	fmt.Printf("Query: %s Args: %s", query, args)
}
```

If we remove the `?` the program will work normally.

My educated guess is that the reason is that the following lines check for the position of the `?` regardless of the existence of an inline comment, i.e. the query isn't being cleaned before bindVars getting evaluated. in `bind.go` line `201`
```go
for i := strings.IndexByte(query[offset:], '?'); i != -1; i = strings.IndexByte(query[offset:], '?') {
		if arg >= len(meta) {
			// if an argument wasn't passed, lets return an error;  this is
			// not actually how database/sql Exec/Query works, but since we are
			// creating an argument list programmatically, we want to be able
			// to catch these programmer errors earlier.
			return "", nil, errors.New("number of bindVars exceeds arguments")
		}
// ....
}

```

c-tonneslan:
Fixes #961.

`sqlx.In()` scans for `?` bind variables using `strings.IndexByte`, which has no concept of SQL syntax. A `?` inside a `--` line comment, a `/* block comment */`, or a single-quoted string literal gets counted as a placeholder, causing a "number of bindVars exceeds arguments" error even when the query is correct.

The fix adds a `nextBindVar` helper that walks the string while tracking comment and string state, returning only positions where `?` is an actual bind variable. The `In` function uses it in place of `IndexByte`.

Added tests covering line comments, block comments, string literals with `?`, and a combined case to make sure real bind variables still expand correctly.

**To reproduce before this fix:**
```go
// Errored with "number of bindVars exceeds arguments"
q, args, err := sqlx.In("SELECT * -- filter by id?\nFROM foo WHERE id IN (?)", []int{1, 2, 3})
```


mrj0:
fyi I have a fix here https://github.com/Vinovest/sqlx/blob/main/bind.go#L191 using sqltokenizer that fixes this and other problems with the parser

PR #985, opened 2026-05-15
named: support PostgreSQL :: cast directly after a named param

### LINKED ISSUE #983: sqlx.Named fails when a named parameter is immediately followed by a PostgreSQL cast
## Version

`github.com/jmoiron/sqlx v1.4.0`

## Summary

`sqlx.Named` appears to misparse PostgreSQL cast syntax when it is attached directly to a named parameter, for example `:boundary::jsonb`.

## Actual Behavior
The call returns:

err=unexpected ':' while reading named param at 16

### Expected Behavior
The query should bind successfully, preserving the PostgreSQL cast, producing something equivalent to:

SELECT ?::jsonb AS boundary, '2026-05-01'::timestamptz AS starts_at;

with one bound argument for boundary.

## Minimal Reproduction

```go
package main

import (
    "fmt"

    "github.com/jmoiron/sqlx"
)

func main() {
    query := "SELECT :boundary::jsonb AS boundary, '2026-05-01'::timestamptz AS starts_at;"
    q, args, err := sqlx.Named(query, map[string]any{
        "boundary": `{"type":"Polygon","coordinates":[]}`,
    })

    fmt.Printf("query=%q\nargs=%#v\nerr=%v\n", q, args, err)
}

c-tonneslan:
Closes #983.

sqlx.Named was bailing with \`unexpected ':' while reading named param at N\` on Postgres queries like:

\`\`\`sql
SELECT :boundary::jsonb FROM ...
\`\`\`

because the colon parser only knew about \`::\` as a way to escape a single colon — it had no story for the cast operator glued directly to the trailing edge of a named parameter.

When \`:\` lands while we've already built up a name, the new code emits the bindvar for that name and then peeks at the next byte. If it's another \`:\`, the whole \`::\` is treated as a Postgres cast and passed through to the rebound query. Otherwise the trailing \`:\` is written as literal text.

The bindvar-emission switch is now factored into an \`emitName\` helper so the new path and the existing end-of-name path stay in sync.

The \`'a::b::c'\` escape behavior inside SQL string literals is intentionally left alone — that case is ambiguous between Postgres cast and escaped colon and isn't what #983 is about.

## Test plan

\`\`\`
go test ./...
\`\`\`

Added a case to TestCompileQuery covering \`:boundary::jsonb, :id::text\` across all bind types.


mrj0:
fyi I have a fix here https://github.com/Vinovest/sqlx/blob/main/bind.go#L191 using sqltokenizer that fixes this and other problems with the parser

itxaiohanglover (commented):
This is a clean approach — extracting emitName avoids the duplication and the test cases cover the main scenarios. The sqltokenizer route would be more invasive for what's ultimately a 2-character lookahead. Any chance this could get reviewed?

PR #987, opened 2026-05-17
fix: preserve PostgreSQL :: casts in Named queries

### LINKED ISSUE #956: Bug in sqlx.Named where ::text is incorrectly processed, leading to syntax erro
**Description:**

I encountered a bug while using sqlx.Named to bind query parameters in SQL queries that contain type casting (::text). The ::text syntax is being modified during the query parameter binding, causing the colon (:) to be removed. This leads to a syntax error in the SQL query.

**Steps to Reproduce:**
Use ::text in a SQL query for type casting in PostgreSQL.
Bind parameters using sqlx.Named.
The resulting query will have the colon (:) removed before text, causing a syntax error.

**Example code:**
```
companyId := 555
query := `SELECT DISTINCT
        t.path::text AS catalog_path
    WHERE t.company_id = :company_id
    FROM table AS t`
params := map[string]interface{}{
    "company_id": companyId,
}
query, args, err := sqlx.Named(query, params)
```

**Resulting Query After sqlx.Named:**
```
SELECT DISTINCT
        t.path:text AS catalog_path
    WHERE t.company_id = :company_id
    FROM table AS t
```

**Error Encountered:**
pq: syntax error at or near ":"

**Expected Behavior:**
The ::text type cast should remain unchanged in the query, and the parameter binding should work as expected.

**Actual Behavior:**
The colon (:) before text is removed, resulting in an invalid SQL query and a syntax error.

wyf027:
## Summary

Fixes #956

`sqlx.Named` was rewriting PostgreSQL type casts like `t.path::text` to `t.path:text`, breaking queries that use `::` for casts.

- Preserve `::` when it follows an identifier character (e.g. `path::text`)
- Preserve `::` when it follows a bound named parameter (e.g. `:boundary::jsonb`)
- Refactor bindvar emission via `appendBindvar` helper

Also adds regression tests for the #956 report and the `:param::type` pattern from #983.

## Test plan

- [x] `go test ./...`
- [x] `TestNamedPostgresCastInIdentifier`
- [x] `TestNamedPostgresCastAfterNamedParam`
- [x] Updated `TestCompileQuery` expectations for string literals containing `::`


mrj0:
fyi I have a whole series of fixes for this using sqltokenizer, see the test https://github.com/Vinovest/sqlx/blob/main/named_test.go#L70

PR #990, opened 2026-06-09
WIP: Select() using iterator

YuriyNasretdinov:
Hi, given the recent proposal for generic methods being accepted into Go (https://github.com/golang/go/issues/77273) I'd like to propose an interator API for Select() which looks like this:

```go
for u, err := db.SelectIter[User]("SELECT * FROM users") {
    if err != nil { ... }
    ...
}
```

This API allows to have convenience of Select() without having to pre-allocate the entire result. The use-case is kinda niche, but without generic methods all other options look much uglier (example: https://github.com/YuriyNasretdinov/sqly).

The code works, however I'm happy to hear your feedback. Marked as WIP since I'd like to add more tests and, crucially, Go 1.27 isn't out yet. The tests do pass on `gotip` though, no GOEXPERIMENT needed


mrj0:
Hi, you might compare this to https://github.com/Vinovest/sqlx

The List function generic makes it easy to avoid preallocation:
```go
people, _ := sqlx.List[Person](db, "SELECT * FROM person")
```

There's also the AllRows function that works with range:
```go
    for person, err := range sqlx.AllRows[Person](rows) {
        if err != nil {
            panic(err)
        }
        fmt.Printf("%#v\n", person)
    }
```

YuriyNasretdinov:
Thanks. Yes, indeed, it is very similar. The main difference is that I'm using generic methods, not just generic functions, and it allows to get a more coherent API. It wasn't possible before go1.27 since generic methods didn't exist.

mrj0:
Yeah I haven't dropped 1.26 yet

PR #992, opened 2026-06-22
fix: add Conn and *Conn to mapperFor and isUnsafe type switches

### LINKED ISSUE #975: The reflectx.Mapper from conn cannot be passed to stmt when creating stmt from conn
I set tagMapFunc，but the reflectx.Mapper from conn cannot be passed to stmt when creating stmt from conn. The mapperFor function lacks handling for database connection (conn) types, which prevents the proper propagation of the reflectx.Mapper to statements created from connections. 
```
func setDBTagMapper(db *sqlx.DB, dbtype DBType) {
	switch dbtype {
	case DBTypeDM, DBTypeORA:
		db.Mapper = reflectx.NewMapperTagFunc("db", nil, strings.ToUpper)
	}
}
my func
stmt, err := conn.PreparexContext(ctx, querySql)
if err != nil {
	t.Error(err)
}
defer stmt.Close()

func PreparexContext(ctx context.Context, p PreparerContext, query string) (*Stmt, error) {
	s, err := p.PrepareContext(ctx, query)
	if err != nil {
		return nil, err
	}
	return &Stmt{Stmt: s, unsafe: isUnsafe(p), Mapper: mapperFor(p)}, err
}
func mapperFor(i interface{}) *reflectx.Mapper {
	switch i := i.(type) {
	case DB:
		return i.Mapper
	case *DB:
		return i.Mapper
	case Tx:
		return i.Mapper
	case *Tx:
		return i.Mapper
	default:
		return mapper()
	}
}
```

xbrxr03:
## Problem

`mapperFor()` and `isUnsafe()` only handle `DB`, `*DB`, `Tx`, and `*Tx` types in their type switches. When a `Stmt` is created from a `*sqlx.Conn`, both functions fall through to the default case:

- `mapperFor(conn)` returns the global mapper instead of `conn.Mapper`
- `isUnsafe(conn)` returns `false` instead of `conn.unsafe`

This silently drops any custom `Mapper` configured on a `Conn` (e.g. via `tagMapFunc`), causing field mapping to use the wrong mapper when creating prepared statements from a connection.

Reported in #975.

## Fix

Add `Conn` and `*Conn` to both type switches:

```go
// mapperFor
case Conn:
    return i.Mapper
case *Conn:
    return i.Mapper

// isUnsafe
case Conn:
    return v.unsafe
case *Conn:
    return v.unsafe
```

## Tests

- `TestMapperForConn`: verifies `mapperFor` returns the correct `Mapper` for `Conn` and `*Conn`, including custom mapper instances
- `TestIsUnsafeConn`: verifies `isUnsafe` returns the correct `unsafe` flag for `Conn` and `*Conn`

Fixes #975


xbrxr03:
Closing this PR due to lack of maintainer engagement. I'm happy to reopen if there's interest. Thanks for the great project.

PR #1002, opened 2026-08-30
Keep PostgreSQL casts attached to named parameters

### LINKED ISSUE #983: sqlx.Named fails when a named parameter is immediately followed by a PostgreSQL cast

official-burak:
`compileNamedQuery` started a named parameter at `:boundary` and then treated the first colon of a following PostgreSQL cast as an unexpected `:` inside that name, so queries like `:boundary::jsonb` never compiled.

After a completed name, `::` is now kept as a type cast (`?::jsonb`, `$1::jsonb`, `@p1::jsonb`). The existing `::` escape (a doubled colon that is not attached to a name, including `:=`) is unchanged. A lone extra colon inside a name (`:first:name`) still errors.

Fixes https://github.com/jmoiron/sqlx/issues/983


Voyagerroc-Lab (approved):
The parser now emits the named bind before preserving an adjacent PostgreSQL :: cast, and the new cases check QUESTION, DOLLAR, AT, and NAMED rebinding plus multiple parameters. The added guard keeps the single-colon-in-name error intact. Review performed with OpenAI Codex at the account owners request.

